PR #14, opened 2019-10-18
Fix Travis Builds Failing

kirstyburgoine:
The second pass at getting the Travis CI builds to pass.

Tested by adding the branch `feature/7-coding-standards` to the list of branches to build. Build successfully passed

Screenshot 2019-10-18 at 14 27 24

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ before_script:
     fi
   - |
     if [[ "$WP_TRAVISCI" == "phpcs" ]] ; then
-      composer global require wp-coding-standards/wpcs
+      composer global require wp-coding-standards/wpcs phpcompatibility/phpcompatibility-wp
       phpcs --config-set installed_paths $HOME/.composer/vendor/phpcompatibility/php-compatibility,$HOME/.composer/vendor/phpcompatibility/phpcompatibility-paragonie,$HOME/.composer/vendor/phpcompatibility/phpcompatibility-wp,$HOME/.composer/vendor/wp-coding-standards/wpcs
     fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ notifications:
 branches:
   only:
     - master
+    - feature/7-coding-standards
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -49,8 +49,8 @@ before_script:
     fi
   - |
     if [[ "$WP_TRAVISCI" == "phpcs" ]] ; then
-      composer global require wp-coding-standards/wpcs
-      phpcs --config-set installed_paths $HOME/.composer/vendor/wp-coding-standards/wpcs
+      composer global require wp-coding-standards/wpcs phpcompatibility/phpcompatibility-wp
```
joemcgill, line 52: We should include a composer.json file in the plugin that has everything we need as a dependency to make sure that the same sniffs we run locally will be run in Travis.

tests/bootstrap.php
```diff
@@ -5,27 +5,27 @@
...
 }
 
 // Give access to tests_add_filter() function.
-require_once $_tests_dir . '/includes/functions.php';
+require_once $core_sitemaps_tests_dir . '/includes/functions.php';
 
 /**
  * Manually load the plugin being tested.
  */
-function _manually_load_plugin() {
```
joemcgill, line 25: I'm not sure that any of these changes in this file are necessary, but if you're changing this function name, you'll also need to change it in the `tests_add_filter()` callback below.

joemcgill (changes requested):

tests/bootstrap.php
kirstyburgoine: @joemcgill These were changed because they are what were failing the sniffs:

Screenshot 2019-10-18 at 14 45 34

Should we be changing the rule in `phpcs.xml.dist` instead to get around this?

joemcgill: Oh interesting. Yeah, that makes sense to me, in case this got installed with another plugin declaring the same symbols.

.travis.yml
kirstyburgoine: @joemcgill I will wait until https://github.com/humanmade/core-sitemaps/pull/15 has been merged before updating this as a `composer.json` file has been created as part of that PR

joemcgill (approved):
Updates look good to me.

PR #26, opened 2019-10-21
Issue & PR templates

kirstyburgoine:
## Description

**Templates for issue creation.** 
Although the motivation for these is for outside contributors to follow a specific format, templates have been created for all of the task labels so that we use them internally as well. Only the task `Spike` doesn't have a template because we pretty much have all of the Spike's outlined now, and it's not expected that third parties would add any more of these.

These are based on the Github defaults with only a few small changes depending on task type.

**Pull request template**
Default pull request template for all types of work.


svandragt:
What is the difference between  a bug-report and a type-bug report in your PR? I don't understand why it's duplicated.

kirstyburgoine:
@svandragt Ah, nothing. Github created this PR automatically and I wasn't quick enough to change it to draft. I still have more to do here

joemcgill (approved):
I think these are a great starting point. Let's give it a go and make changes based on real usage.

PR #28, opened 2019-10-25
[WIP] 18: Posts Sitemaps

### LINKED ISSUE #18: Posts Sitemap
The Sitemap must:

Begin with an opening <urlset> tag and end with a closing </urlset> tag.
Specify the namespace (protocol standard) within the <urlset> tag.
Include a <url> entry for each URL, as a parent XML tag.
Include a <loc> child entry for each <url> parent tag.

https://www.sitemaps.org/protocol.html

Implementation should be informed by performance considerations made previously in #1

Should paginate itself sequentially.

Not included:
Caching
Filters
Post status filters

svandragt:
### Issue Number
Fixes #18.

### Description
A first iteration, that adds content for a first post-type, in this case Posts Sitemaps. It introduces the following draft concepts.

* Sitemap Bucket: up to 2000 sitemap entries, it's data saved into a custom post type as a JSON encoded associative array.
* Sitemap Page: Artificial concept of up to 25 buckets.

The above will probably be simplified in the near future, as having Sitemap Pages with up to 2000 entries scales up to 100m posts by itself (50k * 2k).

It adds a filter for custom post types to hook into the sitemap generation `core_sitemaps_register_post_types`. Sitemap output can be triggered by calling `core_sitemaps_type_post_render()`.

It handles post updates and deletion, creating/updating the sitemap bucket where needed. Missing tests just now as we refine the design.

Code has been commented with placeholder TODO's and FIXME.

image

### Type of change

### Steps to test

1. Create and update posts. 
2. Open twentynineteen's index.php and add `core_sitemaps_type_post_render();die();` before the call to get_header().
3. view source on the homepage.

### Acceptance criteris
- [ ] My changes generate no new warnings. (2 warnings due to placeholder functions).
- [ ] I have added test instructions that prove my fix is effective or that my feature works. (code is subject to change)


composer.json
```diff
@@ -38,10 +41,10 @@
       "vagrant ssh -c \"cd /vagrant/content/plugins/core-sitemaps && WP_TESTS_DB_PASS=password composer run test:phpunit\""
     ],
     "test:phpcs": [
-      "vendor/bin/phpcs -nps --colors --report-code --report-summary --report-width=80 ."
+      "phpcs"
```
joemcgill: I'd expect us to use a vendor version here to ensure everyone is using the same version of phpcs, probably something like `vendor/squizlabs/php_codesniffer/bin/phpcs --standard=./phpcs.xml.dist`.

composer.json
```diff
@@ -8,7 +8,7 @@
     "sort-packages": true,
     "autoloader-suffix": "csm",
     "platform": {
-      "php": "7.3"
+      "php": "5.6"
```
joemcgill: Let's make these composer changes as a separate PR in order to keep the history more clean.

inc/page.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * Each sitemaps has total posts / 50,000 pages.
```
joemcgill, line 3: This is a good idea for packing more URLs into a single page but is probably unnecessary for a first run. Let's start simple by mapping each sitemap page URL to a single bucket of 2000 posts.

inc/type-post.php
```diff
@@ -0,0 +1,94 @@
+<?php
+/**
+ * Posts Sitemap (for post-type posts).
+ */
+
+defined( 'ABSPATH' ) || die();
+
+// Register the a sitemap for the post post-type.
+add_filter( 'core_sitemaps_register_post_types', static function ( $post_types ) {
+	$post_types['post'] = 'core_sitemaps_type_post_register';
```
joemcgill, line 10: This is an ok to get an initial idea, but I think that eventually an object oriented approach that uses a simplified [Registry pattern](https://designpatternsphp.readthedocs.io/en/latest/Structural/Registry/README.html) could be really useful here so we could register different Sitemap classes for each object type. [Yoast does something similar here](https://github.com/Yoast/wordpress-seo/blob/trunk/inc/sitemaps/class-sitemaps.php#L122-L142) which is a nice approach.

core-sitemaps.php
```diff
@@ -17,4 +17,50 @@
...
+}
+
+add_action( 'init', 'core_sitemaps_init', 10 );
+
+/**
+ * Provides the `core_sitemaps_register_post_types` filter to register post types for inclusion in the sitemap.
+ *
+ * @return array Associative array.  Key is the post-type name; Value is a registration callback function.
+ */
+function core_sitemaps_registered_post_types() {
```
joemcgill, line 50: This is an ok starting point, but I think that eventually an object oriented approach that uses a simplified [Registry pattern](https://designpatternsphp.readthedocs.io/en/latest/Structural/Registry/README.html) could be really useful here so we could register different Sitemap classes for each object type. [Yoast does something similar here](https://github.com/Yoast/wordpress-seo/blob/trunk/inc/sitemaps/class-sitemaps.php#L122-L142) which is a nice approach.

inc/bucket.php
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/**
+ * Calculate the sitemap bucket number the post belongs to.
+ *
+ * @param int $post_id Post ID.
+ *
+ * @return int Sitemap Page pagination number.
+ */
+function core_sitemaps_page_calculate_bucket_num( $post_id ) {
```
joemcgill, line 35: I don't think we'll be able to rely on this approach for identifying which bucket a post belongs to. For example, imagine a site had 2000 `page` posts published before publishing a `post` post type. In that case, the first `post` would have an ID that is > 2000, but this would assume that the post should be located in the second bucket, which would be an incorrect assumption.

inc/bucket.php
```diff
@@ -0,0 +1,135 @@
...
+		// Range query.
+		$bucket_meta[] = array(
+			'key'     => 'bucket_num',
+			'value'   => array( $start_bucket, $start_bucket + $max_buckets - 1 ),
+			'type'    => 'numeric',
+			'compare' => 'BETWEEN',
+		);
+	}
+
+	$query_result = $page_query->query(
```
joemcgill, line 78: We'll want to avoid meta queries that are based on meta values, because they are generally slow. We'll probably need to get all `post_meta` rows that use the key of our buckets and loop through each of them to find the one that the post ID we're looking for is in and then map that to the bucket (CPT) ID.

inc/bucket.php
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * @param WP_Post $post Post object.
+ * @param int     $bucket_num Sitemap bucket number.
+ *
+ * @return int|WP_Error @see wp_update_post()
+ */
+function core_sitemaps_bucket_insert( $post, $bucket_num ) {
+	$args = array(
+		'post_type'    => CORE_SITEMAPS_CPT_BUCKET,
+		'post_content' => wp_json_encode(
```
joemcgill, line 99: I really like this idea for being able to update a sitemap node without rebuilding the whole bucket in order to update/remove one post. We may need to implement post locking so that two updates don't trample each other based on race conditions, but we can leave that for later.

joemcgill (commented):
I can see a lot of really good thinking has gone into solving some of the potential performance concerns here, but I think it might be good to step back and start by implementing a basic sitemap page builder that builds out the XML file for all posts in the post type and defer the performance mechanism to a second step.

For now, I think we want to be able to hit a rewrite endpoint like `/sitemap_post.xml` or `/sitemap_post-2.xml` and be able to see the expected sitemap file XML.

composer.json
svandragt: The test:phpcs command is effectively unchanged compared to before this commit, all I've done is move all the parameters inside the phpcs.xml.dist file, which is the file called by phpcs when none is specified.

I simplified the configuration based on the best practices suggested in https://dev.hmn.md/2019/10/10/making-the-best-of-config/.

This can be verified by changing the line  to `which phpcs` it displays _/Users/sander/dev/_hm/core-sitemaps/content/plugins/core-sitemaps/vendor/bin/phpcs_  as for composer this is an equivalent command.

inc/bucket.php
svandragt: Yup and I realised this close after submission, I should have remembered that each post type doesn't have its own ID numbering, instead of course there's just the single wp_posts table.

composer.json
svandragt: See https://github.com/GoogleChromeLabs/wp-sitemaps/pull/30

svandragt:
As discussed, the second pass of this will happen in a new branch.

inc/bucket.php
swissspidy: What if each post has its bucket stored as post meta, so if the post is updated/deleted we now exactly which bucket to update.

Each bucket still contains the same list of entries, but at the same time also information about its size (that could be in wp_options even). When creating a new post, we find the first non-full bucket and add the post to it.

svandragt:
I like that idea.
-- 
Sander van Dragt

Senior Web Engineer
Altis | Human Made

svandragt:
I'll keep the branch https://github.com/GoogleChromeLabs/wp-sitemaps/tree/feature/18-posts-sitemap for reference but this PR won't ever be merged so will close it.

PR #29, opened 2019-10-25
16: sitemap index skeleton

kirstyburgoine:
### Issue Number
https://github.com/swissspidy/wp-sitemaps/issues/16

### Description
Basic skeleton for sitemap_index.xml

### Type of change

### Steps to test
- Activate the plugin
- View yourdomain/sitemap_index.xml

### Acceptance criteria


core-sitemaps.php
```diff
@@ -18,3 +18,7 @@
  */
 
 // Your code starts here.
+
+require_once dirname( __FILE__ ) . '/sitemaps-index.php';
```
svandragt: This is the same as the code below but simpler.
```suggestion
require_once __DIR__ . '/sitemaps-index.php';
```

sitemaps-index.php
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Sets up rewrite rule for sitemap_index.
+	 * @todo Additional rewrites will probably need adding to this.
+	 */
+	public function url_rewrites() {
+		global $wp;
+		$wp->add_query_var( 'sitemap' );
+
+		add_rewrite_rule( 'sitemap_index\.xml$', 'index.php?sitemap=sitemap', 'top' );
```
svandragt: Best prepend the regex with ^ as this will avoid accendental matching on paths ending in the xml file. Sitemap convention dictates the main sitemap is found at sitemap.xml, what's the thinking behind _index?
```suggestion
		add_rewrite_rule( '^sitemap\.xml$', 'index.php?sitemap=sitemap', 'top' );
```

sitemaps-index.php
```diff
@@ -0,0 +1,78 @@
...
+		add_action( 'template_include', array( $this, 'output_sitemap' ) );
+	}
+
+	/**
+	 * Sets up rewrite rule for sitemap_index.
+	 * @todo Additional rewrites will probably need adding to this.
+	 */
+	public function url_rewrites() {
+		global $wp;
+		$wp->add_query_var( 'sitemap' );
```
svandragt: Instead of using the global in this way, the following code can be used that has more safety checks around it.
`add_rewrite_tag('%sitemap%','sitemap');`

sitemaps-index.php
```diff
@@ -0,0 +1,78 @@
...
+	 *
+	 * @todo Split this into seperate functions to apply headers, <xml> tag and <sitemapindex> tag if this is an index?
+	 */
+	public function output_sitemap( $sitemap_content ) {
+
+		$sitemap_index = get_query_var( 'sitemap' );
+
+		if ( ! empty( $sitemap_index ) ) {
+
+			header( 'Content-type: text/xml; charset=' );
```
svandragt: Based on the output below the sitemap contains UTF-8. In that case application/xml is the better content-type. see https://stackoverflow.com/questions/3272534/what-content-type-value-should-i-send-for-my-xml-sitemap, and I think this line misses the charset value?
```suggestion
			header( 'Content-type: application/xml; charset=UTF-8' );
```

svandragt (changes requested):
Good code, very clear. I've added feedback based on other sitemap implementations for your consideration. I'll definiately want to adjust the PR for Posts Sitemaps to work in an object orientated manner.

As an aside, we'd want to flush the rewrite rules when they have changed. In my experience this is best done hooking into init, if update_option('WP_SITEMAPS_REWRITE_VERSION') changed, as this avoids unnecessary flushes and race conditions with checking option values.

inc/sitemaps-index.php
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @var string
+	 */
+	protected $sitemap_content = '';
+
+	/**
+	 * Class constructor.
+	 */
+	public function __construct() {
+		add_action( 'init', array( $this, 'url_rewrites' ), 99 );
```
joemcgill: Let's move actions/filters out of the constructor and into a separate `bootstrap()` method. Even though you will find this pattern common in WordPress code, I generally consider it an anti-pattern that should be avoided so that you can instantiate the class without introducing side effects. It also makes your code more testable because you can load your class without hooking into all the WordPress functionality.

joemcgill (changes requested):
This is looking like a good base to start from. Let's get this in and iterate from here. I only have one suggestion about where to hook actions/filters, otherwise, It think it's good to go.

core-sitemaps.php
```diff
@@ -18,3 +18,7 @@
  */
 
 // Your code starts here.
+
+require_once __DIR__ . '/inc/sitemaps-index.php';
```
joemcgill: Let's rename this file to class-sitemaps-index.php, per https://engineering.hmn.md/standards/style/php/#file-naming.

inc/sitemaps-index.php
```diff
@@ -0,0 +1,73 @@
...
+	 * @param string  $sitemap_content Sitemap Links XML.
+	 * @return string
+	 *
+	 * @todo Split this into seperate functions to apply headers, <xml> tag and <sitemapindex> tag if this is an index?
+	 */
+	public function output_sitemap( $sitemap_content ) {
+		$sitemap_index = get_query_var( 'sitemap' );
+
+		if ( ! empty( $sitemap_index ) ) {
+			header( 'Content-type: application/xml; charset=UTF-8' );
```
swissspidy: Should this use `get_option( 'blog_charset' )`?

joemcgill: No, based on the [sitempas.org spec](https://www.sitemaps.org/protocol.html):

> The Sitemap protocol format consists of XML tags. All data values in a Sitemap must be entity-escaped. The file itself must be UTF-8 encoded.

kirstyburgoine:
@googlebot I consent

inc/sitemaps-index.php
swissspidy: Great!

inc/class-sitemaps-index.php
```diff
@@ -0,0 +1,73 @@
...
+	 * @param string  $template The template to return. Either custom XML or default.
+	 * @return string
+	 *
+	 * @todo Review how the sitemap files are built and placed in the root of the site.
+	 * @todo Split this into seperate functions to apply headers, <xml> tag and <sitemapindex> tag if this is an index?
+	 */
+	public function output_sitemap( $template ) {
+		$sitemap_index = get_query_var( 'sitemap' );
+
+		if ( ! empty( $sitemap_index ) ) {
```
joemcgill: Let's print the XML directly here rather than doing a redirect. We'll end up splitting this out anyway, but we should be able to do so without a 301.

inc/class-sitemaps-index.php
```diff
@@ -0,0 +1,73 @@
...
+		return $redirect;
+	}
+
+	/**
+	 * Produce XML to output.
+	 *
+	 * @param string  $template The template to return. Either custom XML or default.
+	 * @return string
+	 *
+	 * @todo Review how the sitemap files are built and placed in the root of the site.
```
joemcgill: Do we need these `todos` anymore? If so, let's open new tickets and remove the comments from the code.

inc/class-sitemaps-index.php
```diff
@@ -0,0 +1,73 @@
...
+		add_filter( 'template_include', array( $this, 'output_sitemap' ) );
+	}
+
+	/**
+	 * Sets up rewrite rule for sitemap_index.
+	 * @todo Additional rewrites will probably need adding to this.
+	 */
+	public function url_rewrites() {
+		add_rewrite_tag( '%sitemap%','sitemap' );
+		add_rewrite_rule( 'sitemap_index\.xml$', 'index.php?sitemap=sitemap', 'top' );
```
joemcgill: Let's change this to `sitemap=index` or `sitemap=sitemap_index` so it's more descriptive when we check for this value later in our code.

joemcgill (changes requested):
Hi @kirstyburgoine, I've got a few last changes and I think this is ready to go.

inc/sitemap_index.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
swissspidy: This feel doesn't seem to be needed?

joemcgill (approved):
Thanks @kirstyburgoine. I like the simplification for now. Looks like it's working.

svandragt (approved):
Great, please merge this!

PR #30, opened 2019-10-28
Tooling refinements

svandragt:
### Description
Tooling refinements and fixes split from #28. 

- Raises the default minimum environment target to PHP 5.6. 
- Move PHPCS / PHPUnit configuration into the xml configuration files (no effective change).
- Improvement in detection of installed code standards through added composer requirement.
- Simplified test pattern matching (no effective change).
- Remove behat configuration remainders.

Added a composer package to correctly detect all installed code sniffer standards. 

### Type of change

### Steps to test
     composer install
     composer local:tests

### Acceptance criteris

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
       "vagrant ssh -c \"cd /vagrant/content/plugins/core-sitemaps && WP_TESTS_DB_PASS=password composer run test:phpunit\""
     ],
     "test:phpcs": [
-      "vendor/bin/phpcs -nps --colors --report-code --report-summary --report-width=80 ."
+      "phpcs"
     ],
     "test:phpunit": [
       "vendor/bin/phpunit --colors=always"
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
     "phpcompatibility/phpcompatibility-wp": "^2.1",
-    "phpunit/phpunit": "<8.0",
+    "phpunit/phpunit": "^5.0 || ^6.0 || ^7.0",
     "roave/security-advisories": "dev-master",
     "roots/wordpress": "5.2.2",
     "wp-cli/wp-cli": "2.2.0",
--- a/composer.json
+++ b/composer.json
@@ -56,6 +56,7 @@
     "oomphinc/composer-installers-extender": "^1.1"
   },
   "require-dev": {
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
     "phpcompatibility/phpcompatibility-wp": "^2.1",
     "phpunit/phpunit": "<6.0",
     "roave/security-advisories": "dev-master",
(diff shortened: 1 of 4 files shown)


svandragt:
The test:phpcs command is effectively unchanged compared to before this commit, all I've done is move all the parameters inside the phpcs.xml.dist file, which is the file called by phpcs when none is specified.

I simplified the configuration based on the best practices suggested in https://dev.hmn.md/2019/10/10/making-the-best-of-config/.

This can be verified by changing the line  to `which phpcs` it displays _[...]plugins/core-sitemaps/vendor/bin/phpcs_  as for composer this is an equivalent command.

kirstyburgoine (approved):
👍

composer.json
```diff
@@ -56,8 +56,9 @@
     "oomphinc/composer-installers-extender": "^1.1"
   },
   "require-dev": {
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
     "phpcompatibility/phpcompatibility-wp": "^2.1",
-    "phpunit/phpunit": "<6.0",
+    "phpunit/phpunit": "<8.0",
```
joemcgill: Will have to check, but we might need to use the same phpunit version that WP Core uses. This is ok for now.

composer.json
```diff
@@ -38,10 +38,10 @@
       "vagrant ssh -c \"cd /vagrant/content/plugins/core-sitemaps && WP_TESTS_DB_PASS=password composer run test:phpunit\""
     ],
     "test:phpcs": [
-      "vendor/bin/phpcs -nps --colors --report-code --report-summary --report-width=80 ."
+      "phpcs"
```
joemcgill, line 41: I'd prefer to use the dependency version to ensure everyone is getting the same output. Should be: `vendor/squizlabs/php_codesniffer/bin/phpcs`.

joemcgill (approved):
Looking good. Update the location of phpcs and I think this is ready to go in. Going to pre-approve.

composer.json
joemcgill: FYI: for me, running `which phpcs` returns the globally installed path in my users directory.

swissspidy: This should not be needed (i.e. the change is correct).

https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands

composer.json
svandragt: I'm told core is going to use: `"phpunit/phpunit": "^5.0 || ^6.0 || ^7.0"` so i'll go with that then.

composer.json
svandragt: > FYI: for me, running `which phpcs` returns the globally installed path in my users directory.

You're probably running it in the terminal @joemcgill , whereas mean to say if that line is changed in the composer.json; ie. line-41: "which phpcs" composer then resolves to `vendor/bin/phpcs`.

joemcgill: Ah yeah, misunderstood what you were suggesting. Wasn't aware that composer would prioritize a vendor version over whatever was globally available. Nice.

PR #33, opened 2019-10-29
18: Post Sitemaps

### LINKED ISSUE #18: Posts Sitemap

svandragt:
### Issue Number
Fixes #18
Supersedes #28  and #31  

### Description
Post sitemap provider, with pagination. 2000 posts per page. The distribution strategy discussed in #39 is outwith the scope of this PR.

image

### Type of change

### Steps to test
- Displays page 1: http://sitemaps.local/sitemap.xml?sitemap=posts 
- Page 2: http://sitemaps.local/sitemap.xml?sitemap=posts&paged=2

### Acceptance criteris
- [] I have added test instructions that prove my fix is effective or that my feature works.

### DIFF
--- a/inc/class-sitemaps-index.php
+++ b/inc/class-sitemaps-index.php
@@ -17,6 +17,14 @@ class Core_Sitemaps_Index {
 		add_action( 'init', array( $this, 'url_rewrites' ), 99 );
 		add_filter( 'redirect_canonical', array( $this, 'redirect_canonical' ) );
 		add_action( 'template_redirect', array( $this, 'output_sitemap' ) );
+
+		$core_sitemaps_posts = new Core_Sitemaps_Posts();
+		add_action( 'init', array( $core_sitemaps_posts, 'url_rewrites' ), 99 );
+		add_filter( 'template_include', array( $core_sitemaps_posts, 'template' ) );
+
+		// Setup all registered sitemap data providers, after all others.
+		$registry = Core_Sitemaps_Registry::instance();
+		add_action( 'init', array( $registry, 'setup_sitemaps' ), 100 );
 	}
 
 	/**
--- a/inc/class-sitemaps-index.php
+++ b/inc/class-sitemaps-index.php
@@ -44,7 +44,7 @@ class Core_Sitemaps_Index {
 	/**
 	 * Produce XML to output.
 	 *
-	 * @return string
+	 * @return void
 	 *
 	 */
 	public function output_sitemap() {
--- a/inc/class-sitemaps-index.php
+++ b/inc/class-sitemaps-index.php
@@ -23,7 +23,6 @@ class Core_Sitemaps_Index {
 	 * Sets up rewrite rule for sitemap_index.
 	 */
 	public function url_rewrites() {
-		add_rewrite_tag( '%sitemap%','sitemap_index' );
 		$registry = Core_Sitemaps_Registry::instance();
 		$registry->add_sitemap( 'sitemap_index', 'sitemap\.xml$' );
 	}
--- a/inc/class-sitemaps-index.php
+++ b/inc/class-sitemaps-index.php
@@ -5,6 +5,17 @@
  *
  */
 class Core_Sitemaps_Index {
+	/**
+	 * @var Core_Sitemaps_Registry object
+	 */
+	public $registry;
+
+	/**
+	 * Core_Sitemaps_Index constructor.
+	 */
+	public function __construct() {
+		$this->registry = Core_Sitemaps_Registry::instance();
+	}
 
 	/**
 	 *
@@ -18,21 +29,16 @@ class Core_Sitemaps_Index {
 		add_filter( 'redirect_canonical', array( $this, 'redirect_canonical' ) );
 		add_action( 'template_redirect', array( $this, 'output_sitemap' ) );
 
+		// FIXME: Move this into a Core_Sitemaps class registration system.
 		$core_sitemaps_posts = new Core_Sitemaps_Posts();
-		add_action( 'init', array( $core_sitemaps_posts, 'url_rewrites' ), 99 );
-		add_filter( 'template_include', array( $core_sitemaps_posts, 'template' ) );
-
-		// Setup all registered sitemap data providers, after all others.
-		$registry = Core_Sitemaps_Registry::instance();
-		add_action( 'init', array( $registry, 'setup_sitemaps' ), 100 );
+		$core_sitemaps_posts->bootstrap();
 	}
 
 	/**
 	 * Sets up rewrite rule for sitemap_index.
 	 */
 	public function url_rewrites() {
-		$registry = Core_Sitemaps_Registry::instance();
-		$registry->add_sitemap( 'sitemap_index', 'sitemap\.xml$' );
+		$this->registry->add_sitemap( 'sitemap_index', 'sitemap\.xml$' );
 	}
 
 	/**
--- a/inc/class-sitemaps-index.php
+++ b/inc/class-sitemaps-index.php
@@ -58,8 +58,6 @@ class Core_Sitemaps_Index {
 	/**
 	 * Produce XML to output.
 	 *
-	 * @return void
-	 *
 	 */
 	public function render_sitemap() {
 		$sitemap_index = get_query_var( 'sitemap' );
--- a/inc/class-sitemaps-index.php
+++ b/inc/class-sitemaps-index.php
@@ -16,19 +16,6 @@ class Core_Sitemaps_Index {
 	protected $sitemap_content = '';
 
 	/**
-	 *
-	 * A helper function to initiate actions, hooks and other features needed.
-	 *
-	 * @uses add_action()
-	 * @uses add_filter()
-	 */
-	public function bootstrap() {
-		add_action( 'init', array( $this, 'url_rewrites' ), 99 );
-		add_filter( 'redirect_canonical', array( $this, 'redirect_canonical' ) );
-		add_filter( 'template_include', array( $this, 'output_sitemap' ) );
-	}
-
-	/**
 	 * Sets up rewrite rule for sitemap_index.
 	 * @todo Additional rewrites will probably need adding to this.
 	 */
--- a/inc/class-sitemaps-index.php
+++ b/inc/class-sitemaps-index.php
@@ -17,11 +17,11 @@ class Core_Sitemaps_Index {
 
 	/**
 	 * Sets up rewrite rule for sitemap_index.
-	 * @todo Additional rewrites will probably need adding to this.
 	 */
 	public function url_rewrites() {
-		add_rewrite_tag( '%sitemap%','sitemap' );
-		add_rewrite_rule( 'sitemap\.xml$', 'index.php?sitemap=sitemap', 'top' );
+		add_rewrite_tag( '%sitemap%', 'sitemap' );
+		$registry = Core_Sitemaps_Registry::instance();
+		$registry->add_sitemap( 'sitemap', '^sitemap\.xml$' );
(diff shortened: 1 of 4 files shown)


svandragt:
@googlebot I consent.

inc/class-sitemaps-index.php
```diff
@@ -17,14 +17,22 @@ public function bootstrap() {
 		add_action( 'init', array( $this, 'url_rewrites' ), 99 );
 		add_filter( 'redirect_canonical', array( $this, 'redirect_canonical' ) );
 		add_action( 'template_redirect', array( $this, 'output_sitemap' ) );
+
+		$core_sitemaps_posts = new Core_Sitemaps_Posts();
+		add_action( 'init', array( $core_sitemaps_posts, 'url_rewrites' ), 99 );
```
joemcgill: Let's move all hook callbacks for this sitemap into a `bootstrap()` method inside the `Core_Sitemaps_Post` class and call that method once here instead of adding hooks and filters in the sitemaps index.

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,88 @@
...
+	 */
+	public function get_content_per_page( $page_num = 1 ) {
+		$query = new WP_Query();
+
+		return $query->query(
+			array(
+				'orderby'        => 'ID',
+				'order'          => 'ASC',
+				'post_type'      => 'post',
+				'posts_per_page' => 20 * 100,
```
joemcgill: Let's set our maximum posts_per_page as a plugin level constant for now rather than calculating them during the query.

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,88 @@
...
+
+		if ( 'posts' !== $sitemap ) {
+			return $template;
+		}
+
+		$this->content = $this->get_content_per_page( $paged );
+
+		header( 'Content-type: application/xml; charset=UTF-8' );
+		echo '<?xml version="1.0" encoding="UTF-8" ?>';
+		echo '<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">';
```
joemcgill: For sitemap pages, the main XML node should be:
`<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">`

See: https://www.sitemaps.org/protocol.html

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,88 @@
...
+		$registry = Core_Sitemaps_Registry::instance();
+		$registry->add_sitemap( 'posts', '^sitemap-posts\.xml$' );
+	}
+
+	/**
+	 * Produce XML to output.
+	 *
+	 * @param string $template The template to return. Either custom XML or default.
+	 *
+	 * @return string
```
joemcgill: We're not really returning anything here since we're printing output directly and then exiting, so no return value is needed.

inc/class-sitemaps-index.php
```diff
@@ -17,14 +17,22 @@ public function bootstrap() {
...
+		add_filter( 'template_include', array( $core_sitemaps_posts, 'template' ) );
+
+		// Setup all registered sitemap data providers, after all others.
+		$registry = Core_Sitemaps_Registry::instance();
```
joemcgill: I'd like to instantiate one instance of the registry in the plugin and reference that instance rather than relying on the singleton pattern here. For now, let's make this a public property of this class and instantiate the registry in the constructor, but we might end up moving it to a base `Core_Sitemaps` class in the future.

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,88 @@
...
+	 * Content of the sitemap to output.
+	 *
+	 * @var array
+	 */
+	protected $content = [];
+
+	/**
+	 * Sets up rewrite rule for sitemap_index.
+	 */
+	public function url_rewrites() {
```
joemcgill: Let's fire this on the `core_sitemaps_setup_sitemaps` action so we're registering sitemaps just before the registry adds the rewrites. We should also try name our functions/methods so they describe the action being performed. In this case, something like `register_sitemap()` or `setup_sitemap()` would be more clear.

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/**
+	 * Produce XML to output.
+	 *
+	 * @param string $template The template to return. Either custom XML or default.
+	 *
+	 * @return string
+	 */
+	public function template( $template ) {
```
joemcgill: This method name could also be more clear if it described the action it performs, i.e., `render_sitemap()`.

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,88 @@
...
+		$sitemap = get_query_var( 'sitemap' );
+		$paged   = get_query_var( 'paged' );
+
+		if ( 'posts' !== $sitemap ) {
+			return $template;
+		}
+
+		$this->content = $this->get_content_per_page( $paged );
```
joemcgill: Let's not save this content to a property just yet since we're not referencing it anywhere else and will eventually want this class to support multiple post types, which would cause clashes if we're overwriting the content property for each post type and page.

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/**
+	 * Get content for a page.
+	 *
+	 * @param int $page_num Page of results.
+	 *
+	 * @return int[]|WP_Post[] Query result.
+	 */
+	public function get_content_per_page( $page_num = 1 ) {
```
joemcgill: Eventually, we'll need to support multiple post types with this method. Let's have it accept two parameters: `$post_type` and `$page_num`.

inc/class-sitemaps-index.php
```diff
@@ -44,7 +52,7 @@ public function redirect_canonical( $redirect ) {
 	/**
 	 * Produce XML to output.
 	 *
-	 * @return string
+	 * @return void
```
joemcgill: We can completely remove the `return` value here.

joemcgill (changes requested):
The output for the first page here is working and I like the general direction. I've left a bunch of initial feedback notes. I'm not super worried about us supporting multiple post types in this first run, but we should be able to support demonstrate support for multiple pages before we merge this PR.

inc/class-sitemaps-index.php
svandragt: We need to have discuss an approach where sitemap providers can register themselves without adding any code to other classes, this is where a Core_Sitemap class is needed, and is making things more difficult, so that would be a good next task.

svandragt: If we're not using the index class as the main class, then we should probably move the bootstrap call and the registry instantiation into the main core-sitemaps.php file, later replaced by the Core_Sitemaps class there, and keep it out of Index altogether.

I also think replacing a line that explains where something is hooked into something and what it does with a call to a generic method (bootstrap) is not clearer to reason about.

svandragt: What about this in core_sitemaps.php?
```php
function core_sitemaps_bootstrap() {
	$core_sitemaps_index = new Core_Sitemaps_Index();
	$core_sitemaps_index->bootstrap();

	$core_sitemaps_posts = new Core_Sitemaps_Posts();
	$core_sitemaps_posts->bootstrap();

	// Setup all registered sitemap data providers, after all others.
	$registry = Core_Sitemaps_Registry::instance();
	add_action( 'init', array( $registry, 'setup_sitemaps' ), 100 );
};

core_sitemaps_bootstrap();
```
it avoids adding anything to the global namespace and can be replaced with a class later. It is not compatible with calling the registry property of the Core_Sitemaps_Index class from the Core_Sitemaps_Posts class https://github.com/GoogleChromeLabs/wp-sitemaps/pull/33#discussion_r340357084 though.

inc/class-sitemaps-posts.php
svandragt: However the issue here was I wasn't using the template_redirect filter. Now that I switched to that your comment applies.

inc/class-sitemaps-index.php
svandragt: Nevermind the previous comment, it doesn't matter for now as it will get replaced with a Core_Sitemaps class.

inc/class-sitemaps-index.php
svandragt: No, wee need to return $template if the sitemap get variable isn't posts, otherwise the rest of the site goes blank.

svandragt:
@joemcgill All feedback addressed. Also I think you might need to consent again on this PR.

inc/class-sitemaps-posts.php
svandragt: I've also adjusted the method in the index class for consistency.

inc/class-sitemaps-posts.php
svandragt: Done the same on the index class.

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,94 @@
...
+	public function __construct() {
+		$this->registry = Core_Sitemaps_Registry::instance();
+	}
+
+	/**
+	 * Bootstrapping the filters.
+	 */
+	public function bootstrap() {
+		add_action( 'core_sitemaps_setup_sitemaps', array( $this, 'register_sitemap' ), 99 );
+		add_filter( 'template_redirect', array( $this, 'render_sitemap' ) );
```
joemcgill: This should be `add_action` for `template_redirect` rather than `add_filter`.

inc/class-sitemaps-posts.php
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	/**
+	 * Sets up rewrite rule for sitemap_index.
+	 */
+	public function register_sitemap() {
+		$this->registry->add_sitemap( 'posts', '^sitemap-posts\.xml$' );
```
kirstyburgoine, line 32: Do we also need to store a date modified value as an $arg here as well? On the sitemaps index we need to display the `<loc>` and `<lastmod>` values

joemcgill: Eventually, we'll need to be able to get the lastmod from each sitemap page, but let's not worry about that here. For now, just being able to see the sitemap rendered out will be fine.

joemcgill (changes requested):
This is looking good Sander, I can see `sitemap-post.xml` rendering correctly now. I've got one small comment and I think this can merge.

joemcgill:
@googlebot I consent.

inc/class-sitemaps-posts.php
svandragt: Added.

inc/class-sitemaps-posts.php
svandragt: There's no link here or mechanism to the page object to get the last modified date from (either way) so currently that's not possible.

svandragt:
@joemcgill remaining feedback addressed. The distribution strategy discussed in #39 is outwith the scope of this PR and will be progressed in another PR. can merge this after approval.

joemcgill (approved):
Looks like this needs merge conflicts resolved, but my feedback is addressed. Pre-approving.

Issue #35, opened 2019-10-30
List sitemap URLs on index


kirstyburgoine:
### Description

**Which feature is your enhancement request related to?**
https://github.com/GoogleChromeLabs/wp-sitemaps/issues/16

**Describe the solution you'd like**
A follow on from creating the base skeleton for the sitemaps index to display the list of sitemaps in the index.

The Sitemap index may not list more than 50,000 Sitemaps and must be no larger than 50MB (52,428,800 bytes) and can be compressed. 

Sitemap index should be able to list the following sitemap types:

- Core Post Types (Pages and Posts)
- Custom Post Types
- Core Taxonomies (Tags and Categories)
- Custom Taxonomies
- Users (Authors)

**Additional context**
https://support.google.com/webmasters/answer/183668?hl=en
https://www.sitemaps.org/protocol.html

### Acceptance Criteria
Add the relevant acceptance criteria here.
- [x] Lists all of the available sitemap types in the index
- [x] Follows sitemaps.org protocol
- [x] Cannot list more than 50,000 sitemaps
- [ ] Cannot be larger than 50mb

Issue #16, opened 2019-10-21
Index Sitemap


mcshanea:
Sitemap index files (to group multiple sitemap files)

You can provide multiple Sitemap files, and they should be referenced by URL in a Sitemap index file. Sitemap index files may not list more than 50,000 Sitemaps and must be no larger than 50MB (52,428,800 bytes) and can be compressed. There will be one Sitemap index file. The XML format of a Sitemap index file is very similar to the XML format of a Sitemap file.

`<?xml version="1.0" encoding="UTF-8"?>
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
   <sitemap>
      <loc>http://www.example.com/sitemap1.xml.gz</loc>
      <lastmod>2004-10-01T18:23:17+00:00</lastmod>
   </sitemap>
   <sitemap>
      <loc>http://www.example.com/sitemap2.xml.gz</loc>
      <lastmod>2005-01-01</lastmod>
   </sitemap>
</sitemapindex>`

XML Sitemaps will be _enabled by default_ making the following content types indexable:

Homepage
Posts page
Core Post Types (Pages and Posts)
Custom Post Types
Core Taxonomies (Tags and Categories)
Custom Taxonomies
Users (Authors)

Sitemap URLs, like all values in your XML files, must be entity escaped.

https://www.sitemaps.org/protocol.html#index

https://support.google.com/webmasters/answer/183668?hl=en

joemcgill:
Initial work closed by #38. New issues are being created for follow-up work, like #60 and #71

PR #38, opened 2019-10-30
#35 List Sitemap URLS in Sitemap Index

kirstyburgoine:
### Issue Number
https://github.com/GoogleChromeLabs/wp-sitemaps/issues/35 Which is a follow on from https://github.com/GoogleChromeLabs/wp-sitemaps/issues/16

### Description
Get the links for all of the registered sitemaps and output them with correct formatting on sitemap.xml. Tasks include:
- Limit the number of URLs to include to 50,000
- Updated `sitemap_index` to `index` so its more inline with other sitemap types that don't start with `sitemap`
- Added a function to build the URL's for each sitemap to the provider class
- Updated robots.txt function to use built URL
- Added additional parameters to the add_sitemap() function in the registry class to store sitemap URLs
- Outputted URLs to sitemap.xml

### Additional Note
This is very much work in progress at the moment. The first steps are to make sure we can grab the correct data from `Core_Sitemaps_Registry` and output it onto the page with correct markup. 

Screenshot 2019-11-04 at 12 47 50

### Type of change

### Steps to test
- Check out branch `feature/35-sitemap-index-list-urls`.
- View `your-local/sitemap.xml` to see the sitemap index with sample links.

### Acceptance criteria

### DIFF
--- a/core-sitemaps.php
+++ b/core-sitemaps.php
@@ -18,6 +18,7 @@
  */
 
 const CORE_SITEMAPS_POSTS_PER_PAGE = 2000;
+const CORE_SITEMAPS_INDEX_MAX      = 50000;
 
 require_once __DIR__ . '/inc/class-sitemaps-index.php';
 require_once __DIR__ . '/inc/class-sitemaps-posts.php';
--- a/core-sitemaps.php
+++ b/core-sitemaps.php
@@ -18,7 +18,7 @@
  */
 
 const CORE_SITEMAPS_POSTS_PER_PAGE = 2000;
-const CORE_SITEMAPS_INDEX_MAX      = 50000;
+const CORE_SITEMAPS_MAX_URLS       = 50000;
 
 require_once __DIR__ . '/inc/class-sitemaps-provider.php';
 require_once __DIR__ . '/inc/class-sitemaps-index.php';
--- a/inc/class-sitemaps-index.php
+++ b/inc/class-sitemaps-index.php
@@ -52,43 +52,6 @@ class Core_Sitemaps_Index extends Core_Sitemaps_Provider {
 	}
 
 	/**
-	 * Get all of the available sitemaps from the registry
-	 * and add to an array.
-	 *
-	 * @todo get_registered_sitemaps() and get_sitemap_urls() are looping through teh array and
-	 * nested array to get at the value for ['route']. There is probably a better way to do
-	 * this than two methods that are almost identical.
-	 *
-	 * @return array $sitemaps_list
-	 */
-	public function get_registered_sitemaps() {
-		$sitemaps_list = array();
-		$sitemaps_all = $this->registry->get_sitemaps();
-
-		foreach ( $sitemaps_all as $sitemaps ) {
-			array_push( $sitemaps_list, $sitemaps );
-		}
-
-		return $sitemaps_list;
-	}
-
-	/**
-	 * Get all of the URLs for the sitemaps and add to an array.
-	 *
-	 * @return array $sitemaps_urls
-	 */
-	public function get_sitemap_urls() {
-		$sitemap_urls = array();
-		$sitemaps_list = $this->get_registered_sitemaps();
-
-		foreach ( $sitemaps_list as $sitemap ) {
-			array_push( $sitemap_urls, $sitemap );
-		}
-
-		return $sitemap_urls;
-	}
-
-	/**
 	 * Add the correct xml to any given url.
 	 *
 	 * @todo This will also need to be updated with the last modified information as well.
@@ -113,7 +76,7 @@ class Core_Sitemaps_Index extends Core_Sitemaps_Provider {
 	 */
 	public function render_sitemap() {
 		$sitemap_index = get_query_var( 'sitemap' );
-		$sitemap_urls = $this->get_sitemap_urls();
+		$sitemaps_urls = $this->registry->get_sitemaps();
 
 		if ( 'index' === $sitemap_index ) {
 			header( 'Content-type: application/xml; charset=UTF-8' );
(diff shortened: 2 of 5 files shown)


kirstyburgoine:
@googlebot I consent

swissspidy:
@joemcgill Now it's just your consent that's missing for the CLA :-)

joemcgill:
@googlebot I consent.

inc/class-sitemaps-index.php
```diff
@@ -27,7 +34,8 @@ public function bootstrap() {
 	 * Sets up rewrite rule for sitemap_index.
 	 */
 	public function register_sitemap() {
-		$this->registry->add_sitemap( 'sitemap_index', 'sitemap\.xml$' );
+		$post_type = 'index';
```
svandragt: This line is unused now we're using post_type.
 I think we might need a task to separate $name from $post_type as there is no custom post type called index, but for the code that we have it's good.

inc/class-sitemaps-index.php
```diff
@@ -45,37 +53,82 @@ public function redirect_canonical( $redirect ) {
...
+	/**
+	 * Produce XML to output.
+	 *
+	 * @todo At the moment this outputs the rewrite rule for each sitemap rather than the URL.
+	 * This will need changing.
+	 *
+	 */
+	public function render_sitemap() {
+		$sitemap_index = get_query_var( 'sitemap' );
+		$sitemap_urls = $this->get_sitemap_urls();
```
svandragt:
```suggestion
		$sitemap_urls = $this->registry->get_sitemaps();
```
If this suggestion is taken the `get_sitemaps()` and `get_registered_sitemaps()` methods can be deleted.

inc/class-sitemaps-index.php
```diff
@@ -45,37 +53,82 @@ public function redirect_canonical( $redirect ) {
...
+	 * @todo This will also need to be updated with the last modified information as well.
 	 *
-	 * @return string the sitemap index url.
+	 * @return string $markup
 	 */
-	public function sitemap_index_url() {
-		global $wp_rewrite;
+	public function get_index_url_markup( $url ) {
+		$markup = '<sitemap>' . "\n";
+		$markup .= '<loc>' . $url . '</loc>' . "\n";
```
svandragt:
```suggestion
		$markup .= '<loc>' . esc_url($url) . '</loc>' . "\n";
```
The $url is potentially unsafe and could contain >< or other characters.

inc/class-sitemaps-posts.php
```diff
@@ -10,7 +10,7 @@ class Core_Sitemaps_Posts extends Core_Sitemaps_Provider {
 	 *
 	 * @var string
 	 */
-	protected $post_type = 'post';
+	protected $post_type = 'posts';
```
svandragt: The name of the post post_type is singular. (https://codex.wordpress.org/Function_Reference/register_post_type, see reserved post types),
but I suspect we are using this as a $name or $key really?

In any case this will break inc/class-sitemaps-posts.php:38

inc/class-sitemaps-registry.php
```diff
@@ -80,7 +82,11 @@ public function remove_sitemap( $name ) {
 	 * @return array List of sitemaps.
 	 */
 	public function get_sitemaps() {
-		return $this->sitemaps;
+		$total_sitemaps = count( $this->sitemaps );
+
+		if ( $total_sitemaps <= CORE_SITEMAPS_INDEX_MAX ) {
+			return $this->sitemaps;
+		}
```
svandragt, line 92: Have you thought about what happens when more than 50k urls are in the sitemap? For example we could use `array_slice()` to return the first 50k. We could actually always call it and simplify the logic here, however I'm not sure how performant that is. https://www.php.net/manual/en/function.array-slice.php experiment with the preserve_keys parameter though.

core-sitemaps.php
```diff
@@ -18,6 +18,7 @@
  */
 
 const CORE_SITEMAPS_POSTS_PER_PAGE = 2000;
+const CORE_SITEMAPS_INDEX_MAX      = 50000;
```
svandragt: As this is also the maximum urls in any sitemap, would there be merit in reflecting that in the name?

svandragt (changes requested):
I think the unclarity around $post_type and my confusion in using it for multiple purposes has made the code more difficult. I would suggest to give each a $name and use that except for where the actual post type is needed.

inc/class-sitemaps-index.php
kirstyburgoine: yeah, sorry I updated this and removed it as you were reviewing 🤦‍♀

inc/class-sitemaps-index.php
kirstyburgoine: yeah, sorry this is another small thing I updated while you were reviewing

inc/class-sitemaps-index.php
kirstyburgoine: Thank you! I sure that would be an easier way to do this but I couldn't get it to work. Probably had a typo I hadn't spotted...

inc/class-sitemaps-posts.php
kirstyburgoine: Good point, I was using this for the name as well, I've updated to use `$name` everywhere I needed this instead

inc/class-sitemaps-registry.php
kirstyburgoine: No, I'll confess I didn't think about that. From what I've seen from other examples, they seem to return false if there are more entries than allowed. This seems a better solution though.

svandragt (approved):
Excellent! ready to merge.

Issue #36, opened 2019-10-30
Main Class


svandragt:
We need to have discuss an approach where sitemap providers can register themselves without adding any code to other classes, this is where potentially a Core_Sitemap class is needed, and is making development of additional Sitemap Providers more difficult, so that would be a good next task.

### Draft acceptance criterea

- [x] Sitemap provider registers an instance of itself without adding any code to the registry class.
- [x]  Index is able to iterate all registered sitemap providers.

_Originally posted by @svandragt in https://github.com/GoogleChromeLabs/wp-sitemaps/pull/33_

joemcgill:
Closed by #42.

PR #42, opened 2019-11-04
#36 Core Sitemaps main class.

svandragt:
### Issue Number
#36 

### Description
Add a Core_Sitemap class with a 'core_sitemaps_register_providers' filter registeres core sitemap providers.

There's a question whether the Core_Sitemaps class and Core_Sitemaps_Registry have the same purpose, but I think to keep the main class as lean as possible, only to use it for setup.

### Type of change

### Steps to test
load the following urls:
http://sitemaps.local/sitemap.xml 
http://sitemaps.local/sitemap-posts.xml

### Acceptance criteris


inc/registration.php
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Register included providers.
+ *
+ * @param Core_Sitemaps_Provider[] $providers List of registered providers.
+ *
+ * @return Core_Sitemaps_Provider[] Updated list.
+ */
+function core_sitemaps_registration( $providers ) {
+	$providers['sitemap-index'] = new Core_Sitemaps_Index();
```
svandragt: The idea here is that  the key could be the url for the sitemap? or alternatively the index can iterate on the providers and ask for a certain method, which is probably a nice refinement of this PR.

svandragt: Addressed through https://github.com/GoogleChromeLabs/wp-sitemaps/pull/42/commits/349e62ece2983955fdc43f49c53ac672b4c1e28f as it's an acceptance criterea of #36

core-sitemaps.php
```diff
@@ -19,11 +21,12 @@
 
 const CORE_SITEMAPS_POSTS_PER_PAGE = 2000;
 
+require_once __DIR__ . '/inc/class-sitemaps.php';
 require_once __DIR__ . '/inc/class-sitemaps-provider.php';
 require_once __DIR__ . '/inc/class-sitemaps-index.php';
 require_once __DIR__ . '/inc/class-sitemaps-pages.php';
 require_once __DIR__ . '/inc/class-sitemaps-posts.php';
 require_once __DIR__ . '/inc/class-sitemaps-registry.php';
+require_once __DIR__ . '/inc/registration.php';
```
kirstyburgoine: I don't fully understand why this is needed separately from class-sitemaps as it contains a single function?

inc/registration.php
kirstyburgoine: Would this make more sense to be `posts` rather than `sitemap-posts` etc. as that is what is being registered for each sitemap? i.e. 
```
	public function register_sitemap() {
		$this->registry->add_sitemap( 'posts', '^sitemap-posts\.xml$' );
	}
```

inc/registration.php
```diff
@@ -0,0 +1,19 @@
...
+	$providers['sitemap-posts'] = new Core_Sitemaps_Posts();
+	$providers['sitemap-pages'] = new Core_Sitemaps_Pages();
```
kirstyburgoine: What's the benefit of having this function on its own here instead of doing this inside the `Core_Sitemaps` class?

inc/class-sitemaps.php
```diff
@@ -0,0 +1,49 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Get registered providers.
+	 *
+	 * @return Core_Sitemaps_Provider[]
+	 */
+	public function get_providers() {
```
kirstyburgoine, line 60: I'm a bit confused about this as it doesn't seem to be called anywhere? Is there intent that this will be used somewhere in future or am I missing where it's used?

kirstyburgoine:
@googlebot I consent

core-sitemaps.php
svandragt: Well the idea was to give an example for others to reuse, but it's not that complex and on reviewing your comment I've changed it so it's bundled with the filter as the default value.

inc/registration.php
svandragt: Well the idea was to give an example for others to reuse, but it's not that complex and on reviewing your comment I've changed it so it's bundled with the filter as the default value.

inc/class-sitemaps-pages.php
```diff
@@ -11,6 +11,13 @@ class Core_Sitemaps_Pages extends Core_Sitemaps_Provider {
 	 * @var string
 	 */
 	protected $post_type = 'page';
+	/**
+	 * Sitemap name
+	 * Used for building sitemap URLs.
+	 *
+	 * @var string
+	 */
+	protected $name = 'pages';
```
swissspidy, line 20: Why can't we just use the actual post type? Seems unneeded to have this extra plural.

inc/class-sitemaps.php
svandragt: It's not used anywhere, however I'm thinking it will be used a lot.

For example say that the Sitemap Index wants to call each of the providers to get the url for their sitemap, then they can get all the providers and call a `get_sitemap_url()` method to retrieve this value.

inc/class-sitemaps.php
```diff
@@ -0,0 +1,56 @@
...
+	public function __construct() {
+		$registry = new Core_Sitemaps_Registry();
+		/**
+		 * Provides a 'core_sitemaps_register_providers' filter which contains a associated array of
+		 * Core_Sitemap_Provider instances to register, with the key passed into it's bootstrap($key) function.
+		 */
+		$this->providers = apply_filters(
+			'core_sitemaps_register_providers',
+			[
+				'index' => new Core_Sitemaps_Index(),
```
swissspidy: I don't think the sitemap index should be in this filterable list. It doesn't seem to make sense to allow disabling the sitemap index.

svandragt: Yeah this highlights really that in our implementation the index is implemented as a fake post-type but it isn't really, due to the amount of shared code.

Perhaps we should not have the index extend  the Core_Sitemaps_Provider class at all either, this would clarify the code in other places as well @kirstyburgoine thoughts? Which is outside the scope of this PR.

svandragt: Fixed with https://github.com/GoogleChromeLabs/wp-sitemaps/pull/42/commits/4336bd320fd2842f639a67284e7f28aa55caeb44

inc/class-sitemaps-pages.php
svandragt: The name is used for rewrite rules and in urls https://github.com/GoogleChromeLabs/wp-sitemaps/blob/91094a7d6af70533ec766799a0d2e60897e00321/inc/class-sitemaps-registry.php#L87-L88 whereas the post_type is used for post-type queries. 

The index class has a name but not a post-type. Will need more thinking.

inc/class-sitemaps.php
swissspidy: I don't like how `Core_Sitemaps_Provider` focuses only on post types either 🙂 I'll create an issue to describe some changes I'd like to see there.

joemcgill:
> There's a question whether the `Core_Sitemaps` class and `Core_Sitemaps_Registry` have the same purpose, but I think to keep the main class as lean as possible, only to use it for setup.

I've been thinking about this as well. I definitely think the registry could be a part of the main Sitemaps class. Let's do this in a separate issue to avoid rabbit holing for now.

inc/class-sitemaps-pages.php
svandragt: I'll create a task to progress the difference (if any) between post-type and names and whether the index should derive from the same class.

kirstyburgoine (approved):

Issue #48, opened 2019-11-06
Combine post type sitemaps


svandragt:
### Description
Code for all post type sitemaps is very similar, so they can be combined into one class.

**Which feature is your enhancement request related to?**
#18 #17 #19 

**Describe the solution you'd like**
Loop through  system post types linked above and cpt then create sitemaps for each.

### Acceptance Criteria
Add the relevant acceptance criteria here.
- [] all acceptance critera of 17-19

swissspidy:
We most likely want to still split up sitemaps of the `post` object type by the respective sub types (post, page, etc.). So the URLs and stuff would need to reflect that.

joemcgill:
Ideally, we're going to end up with three main providers for core, which support the following object types:

1. Post Objects (i.e., posts, pages, CPTs)
2. Taxonomies (i.e., categories, tags, custom Taxonomies)
3. Users (i.e., authors)

Each object provider should include the ability to support multiple sub-types, so combining these two post types into a single provider should set up the architecture that will be repeated by the Taxonomy sitemap provider as well.

PR #50, opened 2019-11-06
FIX: Rename $post_type to $object_type

kirstyburgoine:
### Description
This change is so that we can use the same variable to pass both post types and taxonomies to the `get_content_per_page()` method instead of having to add another argument. 

### Type of change

This needs to be updated everywhere `$post_type` is used otherwise it will break all the things.

### Acceptance criteria
- [ ] My changes generate no new warnings.

### DIFF
--- a/inc/class-sitemaps-pages.php
+++ b/inc/class-sitemaps-pages.php
@@ -10,7 +10,7 @@ class Core_Sitemaps_Pages extends Core_Sitemaps_Provider {
 	 *
 	 * @var string
 	 */
-	protected $post_type = 'page';
+	protected $object_type = 'page';
 	/**
 	 * Sitemap name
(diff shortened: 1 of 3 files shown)


swissspidy:
This is technically incorrect as `page` is not an object type in WordPress. `post`, `taxonomy`, `user`, and `site` are object types. `page` is just a sub-type of the `post` object type.

I'd rather hold off here and just tackle #48.

kirstyburgoine:
@swissspidy This fix is so that we don't have to add additional arguments to already existing functions for them to be able to render both post types and taxonomy types. Could we agree on a different name instead? This small change affects both my branch and Sander's so would prefer not to hold it up if possible?

swissspidy (approved):
👍 Assuming this will be changed anyway with #48

Issue #22, opened 2019-10-21
Custom Taxonomies Sitemap


mcshanea:
As a person or machine consuming the WP XML sitemaps, I should see links to the archive pages for all registered taxonomies that are marked as `public`.

## Acceptance requirements

- [ ] For each taxonomy, a `url` entry should be included for each term which contains a published public post.
- [ ] Each `url` entry should include a `loc` child entry containing the URL for that term's archive page (i.e., `get_term_link()`).
- [ ] Each `url` entry should include a `lastmod` child entry containing the modified date for the most recently updated post related to that term.
- [ ] Once a taxonomy sitemap page reaches 2000 `url` entries, a new sitemap page is created for that taxonomy type.

## Technical details

1. Update the `Core_Sitemaps_Categories` sitemap provider to be `Core_Sitemaps_Taxonomies` and add support for generating sitemaps for any public taxonomy.
2. The `$route` property should be updated to reference any taxonomy type, like `'^sitemap-taxonomies-([^-]+)-?([0-9]+)?\.xml$';` so that the first match group becomes the taxonomy type and the second match group becomes the paged parameter.
3. The `get_url_list()` method should be updated to use the taxonomy type parsed from the sitemap URL rather than hard coded.
4. The filter in `get_url_list()` should be renamed `core_sitemaps_taxonomies_url_list` and pass the taxonomy type as the second parameter.
5. The `render_sitemap` method should check to see if the specified taxonomy is public before rendering the sitemap and should return a 404 if the taxonomy doesn't exist or isn't public.

joemcgill:
I've updated the issue description with more specific details on how to implement this in a way that should close #20 as well.

joemcgill:
Addressed by #62.

PR #55, opened 2019-11-11
Improve post query performance.

joemcgill:
### Description
This adds three new arguments to the query in `Core_Sitemaps_Provider::get_content_per_page()` to speed up the raw query time:

* `no_found_rows => true` - keeps WP from doing an initial query to count how many posts are in the database for calculating pagination.
* `update_post_term_cache => false` - keeps WP from running an addtional query to warm the post term cache.
* `update_post_meta_cache => false` - keeps WP from running an addtional query to warm the post meta cache.

Adding these arguments reduced the raw query time of a sitemap page on my local environment, containing ~260k posts, from ~2.5 seconds down to ~1.5 seconds.

### Type of change
Please select the relevant options:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Enhancement (improves an existing feature).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Steps to test
Describe the tests required to verify your changes.
Provide instructions so the PR Tester can check functionality and also list any relevant details and / or dependancies required for your tests.

### Acceptance criteris
- [X] My code follows WordPress coding standards.
- [X] I have performed a self-review of my own code.
- [ ] If the changes are visual, I have cross browser / device tested.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [X] My changes generate no new warnings.
- [ ] I have added test instructions that prove my fix is effective or that my feature works.

PR #62, opened 2019-11-12
#22: Taxonomy Sitemaps

kirstyburgoine:
### Issue Number
https://github.com/GoogleChromeLabs/wp-sitemaps/issues/22

### Description
Refactors the categories sitemaps work to generate sitemaps for all taxonomies.

**Tags**
Screenshot 2019-11-13 at 17 31 27

**Categories**
Screenshot 2019-11-13 at 17 31 41

### Type of change

### Steps to test
- Checkout the branch
- View `your-local/sitemap-taxonomies-category.xml` or `your-local/sitemap-taxonomies-post_tag.xml`

### Acceptance criteria
- [ ] My code follows WordPress coding standards.
- [ ] I have performed a self-review of my own code.
- [ ] My changes generate no new warnings.


inc/class-core-sitemaps-taxonomies.php
```diff
@@ -0,0 +1,153 @@
...
+		// Get all of the taxonomies that are registered.
+		$taxonomies = $this->get_object_sub_types();
+
+		$url_list = array();
+
+		foreach ( $taxonomies as $taxonomy ) {
+			// if the query_var matches a taxonomy name, get the terms for that tax.
+			if ( $type === $taxonomy->name ) {
+
+				$taxonomy_terms = get_terms(
```
kirstyburgoine: This may need changing to use `new WP_Term_Query` instead so that `$paged` parameter can be used still.

inc/class-core-sitemaps-taxonomies.php
```diff
@@ -0,0 +1,153 @@
...
+					array(
+						'fields'     => 'ids',
+						'taxonomy'   => $taxonomy->name,
+						'orderby'    => 'term_order',
+						'hide_empty' => true,
```
kirstyburgoine: also need to use `CORE_SITEMAPS_POSTS_PER_PAGE` in this array

inc/class-core-sitemaps-taxonomies.php
```diff
@@ -0,0 +1,153 @@
...
+					)
+				);
+
+				// Loop through the terms and get the latest post stored in each.
+				foreach ( $taxonomy_terms as $term ) {
+
+					$last_modified = get_posts(
+						array(
+							'tax_query'      => array(
```
kirstyburgoine: This is causing `WARNING | Detected usage of tax_query, possible slow query.` in phpcs so may need changing to something more performant?

kirstyburgoine: Probably need to improve general performance here as well based on: https://github.com/GoogleChromeLabs/wp-sitemaps/pull/55

inc/class-core-sitemaps-taxonomies.php
```diff
@@ -0,0 +1,149 @@
...
+	public function render_sitemap() {
+		global $wp_query;
+
+		$sitemap  = get_query_var( 'sitemap' );
+		$sub_type = get_query_var( 'sub_type' );
+		$paged    = get_query_var( 'paged' );
+
+		$sub_types = $this->get_object_sub_types();
+
+		if ( ! isset( $sub_types[ $sub_type ] ) ) {
```
joemcgill: This will cause all pages on the front end to 404 because this `render_sitemap()` callback gets fired on every `template_redirect` action. All of the logic for handling these sitemaps should happen after the initial `$this->slug === $sitemap` check.

inc/class-core-sitemaps-taxonomies.php
```diff
@@ -0,0 +1,149 @@
...
+	/**
+	 * Get a URL list for a taxonomy sitemap.
+	 *
+	 * @param int $page_num Page of results.
+	 *
+	 * @return array $url_list List of URLs for a sitemap.
+	 */
+	public function get_url_list( $page_num = 1 ) {
+		$type = $this->sub_type; // Find the query_var for sub_type.
+		if ( empty( $type ) ) {
```
joemcgill, line 65: I don't think we ever actually want this fallback behavior. The function should probably fail if this happens for taxonomies.

inc/class-core-sitemaps-taxonomies.php
joemcgill: Yeah, this could end up being really slow. Let's add the following args to the `get_posts()` args to ensure we're only doing a single DB hit for each term:

```
'no_found_rows' => true,
'update_post_term_cache' => false,
'update_post_meta_cache' => false,
```

We may need to revisit this later to make it more performant.

joemcgill (changes requested):

svandragt (commented):
Looking good!

inc/class-core-sitemaps-taxonomies.php
```diff
@@ -0,0 +1,154 @@
...
+			return;
+		}
+
+		// Get all of the taxonomies that are registered.
+		$taxonomies = $this->get_object_sub_types();
+
+		$url_list = array();
+
+		foreach ( $taxonomies as $taxonomy ) {
```
joemcgill: I thought I'd already mentioned this, but I must have forgotten to add the comment. 

I don't think this loop is necessary in this method. I'd prefer to assume that validation for the sub-type is handled in the `render_sitemaps()` method so you can run the term query directly using `$this->sub_type` as the `'taxonomy'` argument.

joemcgill (changes requested):
Changes look good. I somehow left out one bit of review feedback that I'd though I had left yesterday. Once that's addressed, this is ready to go in.

joemcgill (approved):
Looks good. Thanks!

PR #53, opened 2019-11-08
#48 Combine 'post' object type sitemaps

svandragt:
### Issue Number
Fixes #48, Fixes #19, 

### Description
Combine 'post' object type sitemaps. This includes pages, posts, custom post types marked as public (with a filter to remove post types). Excludes attachments. Removes pages provider.

The sitemap's rewrite rule needs to work with object types with and without subtypes, so the rewrite query is now defined within the provider, and overwritten by the post class which requires to capture the sub type in addition to the pagination.

The provided subtype is verified agains the result of a filtered get_post_types(public), excluding attachments, and if valid passed into the get_url_list to get the list of urls for the page.

The pagination out of range check is subbed as it will be implemented after we have a bucket in place (page is > than the available bucket id for the post type)

### Outstanding
- index links for all post types.

### Type of change
Please select the relevant options:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Steps to test
Save permalinks
- visit http://sitemaps.local/sitemap-posts-page.xml (urls without page number default to page 1)
- visit http://sitemaps.local/sitemap-posts-post-1.xml (explicit is ok)
- visit http://sitemaps.local/sitemap-posts-post-2.xml (empty urlset, or second page)

### Acceptance criteris
- [x] My code follows WordPress coding standards.
- [x] I have performed a self-review of my own code.
- [x] If the changes are visual, I have cross browser / device tested.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [x] I have added test instructions that prove my fix is effective or that my feature works.

PR #68, opened 2019-11-13
Enhancement: Index Registration for multi-sitemap providers

svandragt:
Builds on #53 (blocker), so please review the following commits only: https://github.com/GoogleChromeLabs/wp-sitemaps/pull/68/files/1783bd0c4498a865c861cdac9b3579f300108215..3f2a0a3e5fda2e4489ed89a20d988699eea34701

### Issue Number
Enhances #48.

### Description
Supports the registration of sitemap names for providers with multiple sitemaps.
As we are still developing the pagination code, there is no way to iterate the available pages, however urls such as sitemap-posts-page-2.xml are supported still.

The provider has a new method `get_sitemap_slugs` which returns a array of keys. The post provider's array is multi-dimensional with the object_type > sub_type slugs returned.

The checks are made such that it's possible for different object types to have the same object sub type slug. (sitemap-posts-page.xml and sitemap-myctp-page.xml).

Per provider the sitemap slugs get passed into the registry after having their provider attached.
In the renderer all sitemaps are unwound and the filenames assembled from the slugs, separated by '-'.

Other multiple-sitemap providers can add support by adding the following method in their class:
```
	public function get_sitemap_slugs() {
		$sub_types               = $this->get_object_sub_types();
		$sitemaps[ $this->slug ] = array();
		foreach ( $sub_types as $sub_type ) {
			$sitemaps[ $this->slug ][ $sub_type->name ] = true;
		}

		return $sitemaps;
	}
```

Screenshot 2019-11-13 at 16 13 07

### Type of change

### Steps to test
`composer local:flush`
http://sitemaps.local/sitemap.xml - verify all post types are visible.
http://sitemaps.local/sitemap-posts-post.xml
http://sitemaps.local/sitemap-users.xml

### Acceptance criteria


svandragt:
I'm not sure of the naming of `add_sitemap()` which now takes all the sitemap names from a provider, then attaches the provider to each of the names. It might be possible to clean up this function further as the provider itself (apart from verification) is of surprisingly little use to the sitemap now.

joemcgill:
Nice one @svandragt!

This PR highlights two distinct issues that I think we need to address. The first of those is that right now we aren't really registering "sitemaps" when we call `Core_Sitemaps_Registry::add_sitemap()`, but instead are registering sitemap providers, each of which can support multiple sitemap sub-types. Each sub-type then actually generates `max(n objects/2000)` sitemaps for each page. So we need to specify what we _mean_ when we add a sitemap to the system. Are we:

1. Registering a sitemap type, which can implement subtypes however it needs?
2. Registering specific object types, meaning that we really need to change our registry so we're adding one sitemap per object sub-type?

This PR seems to do the latter by changing the function signature to pass an array of subtypes. If we decide to go that way, I don't really like that pattern you've suggested here and would rather we looped through the list of sitemap names and call `add_sitemap()` once for each one. However, I would lean towards sticking with the first option for now and let sitemap providers supply a list of sitemaps it generates which includes both sub-types AND page numbers.

This gets at the second issue that we need to address, which is that each sitemap type and/or provider needs to be able to output a list of sitemap URLs for each paginated sitemap OR the core sitemap system needs to be able to construct paginated URLS itself based on information provided by each registered sitemap about how many pages of objects are present.

For now, I would lean towards leaving the registration system the way we have it, where we're registering a sitemap type, which can support subtypes, and focus on the problem of letting the provider return a list of sitemaps (for subtypes, and pages) that it has available. Having a method that returns sitemaps without paged information doesn't seem to solve the whole issue.

joemcgill:
After writing up my last comment, I was curious how other plugins are handling this issue and seems like WordPress SEO is doing basically the same thing I'm suggesting by adding a method to each provider type that returns all of the sitemap links for the index, including sub-types and pages. See this example: https://github.com/Yoast/wordpress-seo/blob/trunk/inc/sitemaps/class-taxonomy-sitemap-provider.php#L57-L64

I think we should go down that path.

svandragt:
I'll close this PR then in favour of one with the required implementation.

Issue #67, opened 2019-11-13
Add an XML Stylesheet (XSL file)


joemcgill:
### Description

One of the project goals states that developers should be able to provide a custom XML Stylesheet (XSL File). In order to provide this functionality, we should add a default XSL file that can be overridden by plugins or themes.

### Acceptance Criteria

The basic file should display the following:

- [ ] An H1 with text, "XML Sitemap"
- [ ] A description paragraph: "This XML Sitemap is generated by WordPress to make your content more visible for search engines. Learn more about XML sitemaps on [sitemaps.org](https://sitemaps.org/).
- [ ] The sitemap URLs or page URLs should be displayed in a table with URLs in the first column and last modified date in the second column.
- [ ] The page should make use of default WordPress font/color styles.
- [ ] All text should make use of WP i18n functions.
- [ ] There should be filters to override the markup and style of the XSL file.

joemcgill:
Examples to look at:

- Yoast has [a method that outputs the XSL sheet](https://github.com/Yoast/wordpress-seo/blob/trunk/inc/sitemaps/class-sitemaps.php#L408-L439) which includes an [XSL template file](https://github.com/Yoast/wordpress-seo/blob/trunk/css/main-sitemap.xsl).
- Jetpack catches [the request during routing](https://github.com/Automattic/jetpack/blob/master/modules/sitemaps/sitemaps.php#L267-L273) and calls on methods of [a class designed specifically for handling XML styling](https://github.com/Automattic/jetpack/blob/master/modules/sitemaps/sitemap-stylist.php).

Both approaches get the job done. I would advocate for an initial approach that is fairly light as long as it meets the criteria.

PR #75, opened 2019-11-18
#67 Add an XML Stylesheet

kirstyburgoine:
### Issue Number
https://github.com/GoogleChromeLabs/wp-sitemaps/issues/67 - Add an XML Stylesheet (XSL file)
This also addresses https://github.com/GoogleChromeLabs/wp-sitemaps/issues/77  - Make CSS for XSL file filterable

### Description
Adds an .xsl stylesheets file to style all of the sitemaps 

### Type of change

### Steps to test
View any of the stylesheets listed below to see them laid out in a table with some basic styles applied
- http://your-local/sitemap.xml
- http://your-local/sitemap-posts-post.xml
- http://your-local/sitemap-taxonomies-category.xml
- http://your-local/sitemap-users.xml
(You may need to flush your redirects locally)

### Screenshots

**Sitemap Index**

Screenshot 2019-11-21 at 17 35 14

**Pages Sitemap**

Screenshot 2019-11-21 at 17 35 00

### Acceptance criteria


inc/class-core-sitemaps-renderer.php
```diff
@@ -48,6 +72,13 @@ public function render_index( $sitemaps ) {
 		// All output is escaped within the addChild method calls.
 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 		echo $sitemap_index->asXML();
+
+		/**
```
joemcgill: I don't think this is the right location for this filter. Did you mean for it to be in `get_sitemap_stylesheet_url()` instead?

inc/class-core-sitemaps-renderer.php
```diff
@@ -75,5 +106,12 @@ public function render_sitemap( $url_list ) {
 		// All output is escaped within the addChild method calls.
 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 		echo $urlset->asXML();
+
+		/**
```
joemcgill: Same as above. You want to put the filter inside the function that returns the stylesheet location rather than here. This function already echoes the created XML so it doesn't really return anything that would be useful to filter.

inc/class-core-sitemaps.php
```diff
@@ -96,6 +97,13 @@ public function setup_sitemaps() {
 		}
 	}
 
+	/**
+	 * Provide rewrite for the xsl stylesheet.
+	 */
+	public function xsl_stylesheet_rewrite() {
```
joemcgill: Ah, I see what's happening here.

The rewrite rule will need to direct to something like `index.php?sitemap-xsl=1`, and then you can add a method to the `Core_Sitemaps` class that works similar to how `Core_Sitemaps_Index::render_sitemap()` works, where you hook to the `template_redirect` action with a function that looks for the existence of the `sitemap-xsl` param—which you'll need to add an `add_rewrite_tag` for—and then include your sitemap.xsl.php file programmatically and `exit`.

inc/class-core-sitemaps-stylesheet.php
```diff
@@ -0,0 +1,156 @@
...
+			header( 'Content-type: application/xml; charset=UTF-8' );
+
+			$xsl = $this->stylesheet_xsl();
+
+			/**
+			 * Filter the content of the sitemap stylesheet.
+			 *
+			 * @param string $xsl Full content for the xml stylesheet.
+			 */
+			return apply_filters( 'core_sitemaps_stylesheet_content', $xsl );
```
joemcgill: Instead of returning here, you'll need to echo the XSL and then `exit` so WordPress doesn't continue with it's templating logic.

inc/class-core-sitemaps-stylesheet.php
```diff
@@ -0,0 +1,234 @@
...
+		 * @param string $xsl Full content for the xml stylesheet.
+		 */
+		return apply_filters( 'core_sitemaps_stylesheet_content', $xsl_content );
+	}
+
+
+	/**
+	 * Returns the escaped xsl for the index sitemaps.
+	 */
+	public function stylesheet_index_xsl() {
```
kirstyburgoine, line 118: In commit https://github.com/GoogleChromeLabs/wp-sitemaps/pull/75/commits/ae5ad7759f8d82974de0723dd8f6dd8359c04cd4 I combined both of these so that there was only one output and one filter, and the content within changed based on the query_var. I later realised though that if anyone was going to filter this content they would also need to add in checks for whether it was the index sitemap or not so decided it was better to keep them completely seperate

inc/class-core-sitemaps-stylesheet.php
```diff
@@ -0,0 +1,234 @@
...
+	 * Renders the xsl stylesheet depending on whether its the sitemap index or not.
+	 */
+	public function render_stylesheet() {
+		$stylesheet_query = get_query_var( 'stylesheet' );
+
+		if ( ! empty( $stylesheet_query ) ) {
+			header( 'Content-type: application/xml; charset=UTF-8' );
+
+			if ( 'xsl' === $stylesheet_query ) {
+				// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- All content escaped below.
```
joemcgill, line 24: 👍

inc/class-core-sitemaps.php
```diff
@@ -96,6 +98,27 @@ public function setup_sitemaps() {
...
+		add_action( 'template_redirect', array( $stylesheet, 'render_stylesheet' ) );
+	}
+
+	/**
+	 * Provide rewrite for the sitemap index xsl stylesheet.
+	 */
+	public function index_xsl_stylesheet_rewrite() {
+		add_rewrite_rule( '^sitemap-index\.xsl$', 'index.php?stylesheet=index', 'top' );
+
+		$stylesheet = new Core_Sitemaps_Stylesheet();
```
joemcgill, line 108: These two lines are redundant with the ones above in `xsl_stylesheet_rewrite()` above since both are going to fire on the `init` action. I would combine these two methods int a single method that adds rewrite rules for both stylesheet types.

joemcgill (changes requested):
This looks good and is working great now, @kirstyburgoine. Nice work! There's one change I'd like to see with the way rewrites are handled (comment below) and then I think this is good to merge.

joemcgill (approved):

Issue #74, opened 2019-11-16
Make max urls per sitemap filterable


joemcgill:
### Description

**Which feature is your enhancement request related to?**
All sitemaps.

**Describe the solution you'd like**
Currently, the max amount of URLs included in a sitemap are controlled by the `CORE_SITEMAPS_POSTS_PER_PAGE` constant, which gets referenced throughout the code, but particularly in each provider's `get_url_list()` function to limit object queries to a maximum number of results.

Instead, we should add a helper function called something like `get_max_sitemap_urls()` and return the value of the constant after being run through a filter. This way, people can modify the max number of URLs to fit their environment.

Another benefit of doing so is that we can filter this to a low value in our unit tests to test pagination functionality without needing to generate thousands of test objects and associations during our unit testing.

Additionally, the constant should be renamed to a more accurate name, since we're not always using it to limit post objects. I propose `CORE_SITEMAPS_MAX_URLS_PER_SITEMAP`.

### Acceptance Criteria

- [ ] A helper function is added to the functions file that returns the default value for `CORE_SITEMAPS_POSTS_PER_PAGE`.
- [ ] The function contains a filter so the default value can be modified.
- [ ] All instances in the code where `CORE_SITEMAPS_POSTS_PER_PAGE` is used should be replaced by the newly added function.
- [ ] The constant can also be overridden if manually set in `wp-config` or similar.

joemcgill:
@svandragt brought up the point on our call today that we could also allow this constant to be set in `wp-config.php` by checking for `if ( ! defined() )` before setting this constant.

We should allow for that as well, so I'm going to add that as criteria, but would also be good for people to be able to filter this value for specific circumstances without overriding all values, e.g., filtering the URLs/sitemap page for only a specific object type.

PR #72, opened 2019-11-15
Feature/index registration mk2

svandragt:
### Issue Number
2nd iteration following #68, adding to #48.

Fixes #71.

### Description
Supports the registration of sitemap names for providers with 0 or more object sub types.
This includes iterating pages.

Each provider provides the sitemaps that should be available in the index which are object-type, sub-type and page specific.

Third party providers add support by adding `max_num_pages($type=null)` method that returns the total page count for the provided object (sub)type, as this depends on how the type is queried (type is optional for providers without sub types).

**Known issues:**
Some of the page count queries can perhaps be simplified using SELECT COUNT(ID) perhaps. and ~there might not be a need to build up an stub object type if can can just pass the name and simplify it~ (update: simplified), but not looked into those things yet. Because the new methods return a page count, they should be easily switchable when the bucket cpt is ready.

### Screenshots (before and after if applicable)
image

### Type of change
Please select the relevant options:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Enhancement (change which improves an existing feature. E.g., performance improvement, docs update, etc.)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Steps to test
SM-Tests
http://sitemaps.local/sitemap.xml
http://sitemaps.local/sitemap-posts-post.xml
http://sitemaps.local/sitemap-posts-post-1.xml
http://sitemaps.local/sitemap-users.xml
http://sitemaps.local/sitemap-taxonomies-post_format-1.xml
http://sitemaps.local/sitemap-users-1.xml
http://sitemaps.local/sitemap-taxonomies-post_tag-1.xml

### Acceptance criteria
- [x] My code follows WordPress coding standards.
- [x] I have performed a self-review of my own code.
- [x] If the changes are visual, I have cross browser / device tested.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [x] I have added test instructions that prove my fix is effective or that my feature works.

PR #76, opened 2019-11-18
Fix sitemap pagination for taxonomy sitemap pages

joemcgill:
### Issue Number
This is blocked by #72. All changes are in 0ddcb36.

### Description
Previously, the paged value was being discarded when passed to the `WP_Term_Query` when getting URLs for a taxonomy sitemap page. This fixes the arguments passed to the `WP_Term_Query` so the queries return the expected values for each page.

This would be easier to test once #74 is implemented.

### Type of change

### Steps to test
- Manually change the URL of a taxonomy sitemap page to a page number that is higher than the available terms and see that the first page of results are still returned.

### Acceptance criteria

### DIFF
--- a/inc/class-core-sitemaps-taxonomies.php
+++ b/inc/class-core-sitemaps-taxonomies.php
@@ -68,16 +68,16 @@ class Core_Sitemaps_Taxonomies extends Core_Sitemaps_Provider {
 		$offset = ( $page_num - 1 ) * CORE_SITEMAPS_POSTS_PER_PAGE;
 
 		$args = array(
-			'taxonomy'     => $type,
-			'orderby'      => 'term_order',
-			'number'       => CORE_SITEMAPS_POSTS_PER_PAGE,
-			'offset'       => $offset,
-			'hide_empty'   => true,
+			'taxonomy'               => $type,
+			'orderby'                => 'term_order',
+			'number'                 => CORE_SITEMAPS_POSTS_PER_PAGE,
+			'offset'                 => $offset,
+			'hide_empty'             => true,
 			/*
 			 * Limits aren't included in queries when hierarchical is set to true (by default).
 			 * See: https: //github.com/WordPress/WordPress/blob/5.3/wp-includes/class-wp-term-query.php#L558-L567
 			 */
-			'hierarchical' => false,
+			'hierarchical'           => false,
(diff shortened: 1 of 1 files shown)


joemcgill:
CLAs will fix themselves after #72 gets merged. I'll handle CS issues after updating.

inc/class-core-sitemaps-taxonomies.php
```diff
@@ -64,42 +64,52 @@ public function get_url_list( $page_num ) {
...
-			'paged'      => absint( $page_num ),
-			'hide_empty' => true,
+			'taxonomy'     => $type,
+			'orderby'      => 'term_order',
+			'number'       => CORE_SITEMAPS_POSTS_PER_PAGE,
+			'offset'       => $offset,
+			'hide_empty'   => true,
+			/*
+			 * Limits aren't included in queries when hierarchical is set to true (by default).
+			 * See: https: //github.com/WordPress/WordPress/blob/5.3/wp-includes/class-wp-term-query.php#L558-L567
```
svandragt: If we want to use PHPDoc lingo here then use @link here (and a blank line before it haha finally!)

inc/class-core-sitemaps-taxonomies.php
```diff
@@ -75,7 +75,8 @@ public function get_url_list( $page_num ) {
 			'hide_empty'             => true,
 			/*
 			 * Limits aren't included in queries when hierarchical is set to true (by default).
-			 * See: https: //github.com/WordPress/WordPress/blob/5.3/wp-includes/class-wp-term-query.php#L558-L567
+			 *
+			 * @link: https://github.com/WordPress/WordPress/blob/5.3/wp-includes/class-wp-term-query.php#L558-L567
```
joemcgill, line 81: 💖 This is for @svandragt.

kirstyburgoine (approved):
👍 Added one question, but it's not a blocker

joemcgill:
Thanks @kirstyburgoine, good eye on that one.

I had originally removed that param to test out the theory that pagination wasn't working properly because of this bug: https://core.trac.wordpress.org/ticket/41796. However, even though learning about that issue was interesting, that bug was not related to the reason this wasn't working, so I should have put that parameter back in place.

PR #82, opened 2019-11-19
#74 Make max URLs per sitemap values filterable

### LINKED ISSUE #74: Make max urls per sitemap filterable

joemcgill:
### Issue Number
This fixes #74.

### Description
This allows the maximum number of URLs per sitemap to be overridden by setting the value via a constant defined in `wp-config.php` or by filtering the value during runtime.

**Main changes included:**

- Rename constant for maximum sitemaps to `CORE_SITEMAPS_MAX_SITEMAPS`. (7d26d12)
- Rename constant for limiting URLs in a sitemap to `CORE_SITEMAPS_MAX_URLS`. (2862efe)
- Add `core_sitemaps_get_max_urls()` function and with `core_sitemaps_max_urls` filter to modify max URL per sitemap value by sitemap type. (3163240)
- Add unit tests for `core_sitemaps_get_max_urls()` (9d00ed1)

### Type of change

### Steps to test
1. Add `define( 'CORE_SITEMAPS_MAX_URLS', 100 );` to your `wp-config.php` and visit any sitemap type to see that there are no more than 100 URLs on a sitemap.
2. Add the following filter to an MU plugin or somewhere that will be run by WordPress and see that the number of URLs on each object type has been modified based on type.

```
add_filter( 'core_sitemaps_max_urls', function( $max_urls, $type ) {
    switch ( $type ) {
        case 'posts':
            return 300;
        case 'taxonomies':
            return 50;
        case 'users':
            return 10;
        default:
            return $max_urls;
    }
}, 10, 2 );
```

### Acceptance criteria


joemcgill:
Interesting. Looks like the unit tests are failing because the `core_sitemaps_get_max_urls()` function isn't defined. We're not bootstrapping our plugin during the test run. Will look into this tomorrow.

kirstyburgoine (approved):

joemcgill:
🙌 Unit tests are fixed. Merging.

Issue #80, opened 2019-11-19
Add blog homepage to the index sitemap.


joemcgill:
### Description

[TK full description]

When the front page is a blog index, it should be included in the sitemap index or should be included on the first page of the pages sitemap.

**Describe the solution you'd like**
A clear and concise description of what the enhancement is and what you want to happen.

**Additional context**
Let's review how other plugins are handling this.

### Acceptance Criteria
Add the relevant acceptance criteria here.
- []
- []
- []

joemcgill:
Fixed by #89.

PR #89, opened 2019-11-22
Feature/80 add homepage

kirstyburgoine:
### Issue Number
https://github.com/GoogleChromeLabs/wp-sitemaps/issues/80 

### Description
If the homepage has been set to show latest posts, then the main homepage URL displays on the pages sitemap with a lastmod date of the latest post on the site. 

This isn't needed if set to display a static page as that page is included within the pages sitemap automatically, and displays the lastmod date for that page.

**Your homepage displays latest posts**

Screenshot 2019-11-22 at 15 58 13

**Your homepage displays a static page**

Screenshot 2019-11-22 at 15 58 35

### Type of change

### Steps to test
- View the sitemap index locally
- Log in to the admin in another tab and go to Settings > Reading.
- Set your WordPress install to "your homepage displays latest posts" to see the home url display on the sitemap index.

### Acceptance criteria


swissspidy:
That doesn‘t sound valid to me. Per https://www.sitemaps.org/protocol.html#index, the index must only point to other sitemaps, and not actual sitemap entries. Can you double check this?

kirstyburgoine:
Added WIP to this as I have some follow up work to do here so it's not ready for review, but can't seem to make it a draft again. I will update when it's ready again

kirstyburgoine:
Ok, this should be ready for review now :)

inc/class-core-sitemaps-renderer.php
```diff
@@ -113,6 +114,7 @@ public function render_index( $sitemaps ) {
 	 */
 	public function render_sitemap( $url_list ) {
 		global $wp_query;
+		$sub_type = get_query_var( 'sub_type' );
```
joemcgill: After your latest changes, this variable is no longer used and can be removed.

inc/class-core-sitemaps-provider.php
```diff
@@ -111,6 +111,27 @@ public function get_url_list( $page_num ) {
 			);
 		}
 
+		// Show a URL for the homepage in the pages sitemap if the reading settings are set to display latest posts.
```
joemcgill: We need to add a check to make sure this is on the first page only. Otherwise, the homepage URL will show up on every sitemap where `$post_type === 'page'`.

Nitpick, but I'd prefer to see this happen before the rest of the URLs are added, so that the homepage URL is the first URL in the list, rather than the last.

joemcgill (dismissed):

joemcgill:
@svandragt can you give this a look now that @kirstyburgoine and I have both made changes here?

inc/class-core-sitemaps-provider.php
```diff
@@ -136,6 +136,31 @@ public function get_url_list( $page_num, $type = '' ) {
...
+		/*
+		 * Add a URL for the homepage in the pages sitemap.
+		 * Shows only on the first page if the reading settings are set to display latest posts.
+		 */
+		if ( 'page' === $type && 1 === $page_num && 'posts' === get_option( 'show_on_front' ) ) {
+			// Assumes the homepage last modified date is the same as the most recent post.
+			$last_modified = get_posts(
+				array(
+					'post_type'              => 'post',
+					'posts_per_page'         => '1',
```
goldenapples: Is there a reason this value is being passed as a string? The WP_Query docs expect an integer here.

inc/class-core-sitemaps-provider.php
```diff
@@ -136,6 +136,31 @@ public function get_url_list( $page_num, $type = '' ) {
...
+					'orderby'                => 'date',
+					'order'                  => 'DESC',
+					'no_found_rows'          => true,
```
goldenapples, line 148: I'm not 100% sure that this query will always return the most recent modified date. I'm thinking of various combinations of query filters that could affect the actual posts that make up the homepage, and as a result the most recently modified post on the home index might not be the post which this query returns.

I don't know if its possible to account for every possible way an archive page might be structured, but to get a more accurate representation of the last modified date of the home index, I would think we'd need to set up a query that looks like the WP query on the home url, and take the most recent modified date from the posts in that query.

This might be overkill for the actual acceptance criteria though... just putting the question out there since it wasn't clear to me what actually needs to be done as part of this issue.

goldenapples (commented):
Left a couple questions on the query here. The code looks good, though it doesn't look like this will necessarily give an accurate last modified date.

inc/class-core-sitemaps-provider.php
joemcgill: Yeah, I agree that this is a very opinionated assumption, but does cover the default use case for WordPress wherein you've set the homepage as your page for posts (i.e. blog front page), which I think is fine for an initial PoC.

inc/class-core-sitemaps-provider.php
joemcgill: Good catch.

joemcgill: Simplified the args here in 8f596b2.

goldenapples (approved):
Looks good. 

As discussed, it's a little unclear what to send as the exact "last modified" date for index pages. The query here is a decent approximation that should give meaningful values in most use cases.  Approved.

inc/class-core-sitemaps-provider.php
svandragt: As a refinement for a future update, as an untested theory it could be worth investigating if it would work to hook into pre_get_posts and check if it's the main loop, then within these checks and then rewind the loop and get the most recent post from it.

svandragt (approved):
Sorry for not being aware of this sooner, I have left a comment to review but the work itself is good to go.

PR #90, opened 2019-11-27
Add lastmod values to the sitemap index

### LINKED ISSUE #60: `lastmod` entry for sitemaps are hard coded in the index.
**Describe the bug**
Each sitemap in the index should contain a `lastmod` entry that shows when that sitemap was last modified. However, this value is currently hard coded in `Core_Sitemaps_Renderer::render_index()` and never updated.

**To Reproduce**
1. Go to /sitemap.xml
2. Notice that all sitemaps have the same `lastmod` date.

**Expected behavior**
The `lastmod` value of a sitemap should be equal to the most recent `lastmod` value from the `url` entries of that sitemap, or whenever that sitemap was last modified (whichever is later).

joemcgill:
This implements a performant way of showing `lastmod` values for sitemap pages in the sitemap index.

### Issue Number
This will fix #60.

### Description
It's not performant to calculate `lastmod` values dynamically when the site index is rendered, so this approach calculates the `lastmod` values asynchronously. `lastmod` values for each sitemap page is stored in an option with a key of `core_sitemaps_lasmod_{object-type}_{object-subtype}_{page_number}` which are checked whenever the sitemap index is rendered. If no value exists, a single job is scheduled to fill in that value. All values are updated twice daily rather than updating them dynamically each time a post, taxonomy archive, or user is changed.

The main changes include:

- Moves responsibility for storing sitemap pages from the registry to each sitemap provider (7a64d11), which also improves general performance.
- Calculate `lastmod` values asynchronously and store as options (7ee9014, 6b9bab6).
- Update all `lastmod` values via a scheduled task run twice daily (56c3c16, 4da5f94, and other cleanup).

### Type of change

### Steps to test
- Visit the sitemap index for the first time to see that all lastmod values are empty.
- Run `wp cron event list` to see that events have been scheduled for calculating lastmod values.
- Wait for the events to run, or manually run `wp cron event run --all` to clear the scheduled events and see that `lastmod` values have been updated.

### Acceptance criteria

### DIFF
--- a/inc/class-core-sitemaps-index.php
+++ b/inc/class-core-sitemaps-index.php
@@ -75,7 +75,7 @@ class Core_Sitemaps_Index {
 			$sitemaps = array();
 
 			// Build up the list of sitemap pages.
-			foreach( $providers as $provider ) {
+			foreach ( $providers as $provider ) {
 				$sitemaps = array_merge( $sitemaps, $provider->get_sitemap_entries() );
 			}
 
--- a/inc/class-core-sitemaps-index.php
+++ b/inc/class-core-sitemaps-index.php
@@ -70,8 +70,16 @@ class Core_Sitemaps_Index {
 		$sitemap_index = get_query_var( 'sitemap' );
 
 		if ( 'index' === $sitemap_index ) {
-			$sitemaps = core_sitemaps_get_sitemaps();
-			$this->renderer->render_index( array_keys( $sitemaps ) );
+			$providers = core_sitemaps_get_sitemaps();
+
+			$sitemaps = array();
+
+			// Build up the list of sitemap pages.
+			foreach( $providers as $provider ) {
+				$sitemaps = array_merge( $sitemaps, $provider->get_sitemap_entries() );
+			}
+
+			$this->renderer->render_index( $sitemaps );
 			exit;
 		}
 	}
--- a/inc/class-core-sitemaps-index.php
+++ b/inc/class-core-sitemaps-index.php
@@ -74,7 +74,6 @@ class Core_Sitemaps_Index {
 
 			$sitemaps = array();
 
-			// Build up the list of sitemap pages.
 			foreach ( $providers as $provider ) {
 				$sitemaps = array_merge( $sitemaps, $provider->get_sitemap_entries() );
 			}
--- a/inc/class-core-sitemaps-provider.php
+++ b/inc/class-core-sitemaps-provider.php
@@ -279,7 +279,12 @@ class Core_Sitemaps_Provider {
 
 		// If blank, schedule a job.
 		if ( empty( $lastmod ) && ! wp_doing_cron() ) {
-			wp_schedule_single_event( time() + 500, 'core_sitemaps_calculate_lastmod', array( $this->slug, $name, $page ) );
+			$event_args = array( $this->slug, $name, $page );
+
+			// Don't schedule a duplicate job.
+			if ( ! wp_next_scheduled( 'core_sitemaps_calculate_lastmod', $event_args ) ) {
+				wp_schedule_single_event( time() + 500, 'core_sitemaps_calculate_lastmod', $event_args );
+			}
 		}
 
 		return $lastmod;
--- a/inc/class-core-sitemaps-provider.php
+++ b/inc/class-core-sitemaps-provider.php
@@ -207,6 +207,31 @@ class Core_Sitemaps_Provider {
 	}
 
 	/**
+	 * Get data about each sitemap type.
+	 *
+	 * @return array List of sitemap types including object subtype name and number of pages.
+	 */
+	public function get_sitemap_type_data() {
+		$sitemap_data = array();
+
+		$sitemap_types = $this->get_object_sub_types();
+
+		foreach ( $sitemap_types as $type ) {
+			// Handle lists of post-objects.
+			if ( isset( $type->name ) ) {
+				$type = $type->name;
+			}
+
+			$sitemap_data[] = array(
+				'name'   => $type,
+				'pages' => $this->max_num_pages( $type ),
+			);
+		}
+
+		return $sitemap_data;
+	}
+
+	/**
 	 * List of sitemap pages exposed by this provider.
 	 *
 	 * The returned data is used to populate the sitemap entries of the index.
@@ -216,22 +241,13 @@ class Core_Sitemaps_Provider {
 	public function get_sitemap_entries() {
 		$sitemaps = array();
 
-		$sitemap_types = $this->get_object_sub_types();
+		$sitemap_types = $this->get_sitemap_type_data();
 
 		foreach ( $sitemap_types as $type ) {
-			// Handle object names as strings.
-			$name = $type;
 
-			// Handle lists of post-objects.
-			if ( isset( $type->name ) ) {
-				$name = $type->name;
-			}
-
-			$total = $this->max_num_pages( $name );
-
-			for ( $page = 1; $page <= $total; $page ++ ) {
-				$loc        = $this->get_sitemap_url( $name, $page );
-				$lastmod    = $this->get_sitemap_lastmod( $name, $page );
+			for ( $page = 1; $page <= $type['pages']; $page ++ ) {
+				$loc        = $this->get_sitemap_url( $type['name'], $page );
+				$lastmod    = $this->get_sitemap_lastmod( $type['name'], $page );
 				$sitemaps[] = array(
 					'loc'     => $loc,
 					'lastmod' => $lastmod,
@@ -332,21 +348,11 @@ class Core_Sitemaps_Provider {
 	 * Schedules asynchronous tasks to update lastmod entries for all sitemap pages.
 	 */
 	public function update_lastmod_values() {
-		$sitemap_types = $this->get_object_sub_types();
+		$sitemap_types = $this->get_sitemap_type_data();
 
 		foreach ( $sitemap_types as $type ) {
-			// Handle object names as strings.
-			$name = $type;
-
-			// Handle lists of post-objects.
-			if ( isset( $type->name ) ) {
-				$name = $type->name;
-			}
-
-			$total = $this->max_num_pages( $name );
-
-			for ( $page = 1; $page <= $total; $page ++ ) {
-				wp_schedule_single_event( time(), 'core_sitemaps_calculate_lastmod', array( $this->slug, $name, $page ) );
+			for ( $page = 1; $page <= $type['pages']; $page ++ ) {
+				wp_schedule_single_event( time(), 'core_sitemaps_calculate_lastmod', array( $this->slug, $type['name'], $page ) );
 			}
 		}
 	}
--- a/inc/class-core-sitemaps-provider.php
+++ b/inc/class-core-sitemaps-provider.php
@@ -316,20 +316,16 @@ class Core_Sitemaps_Provider {
 			return;
 		}
 
-		$list = $this->get_url_list( $page, $subtype );
+		// Get the list of URLs from this page and sort it by lastmod date.
+		$url_list    = $this->get_url_list( $page, $subtype );
+		$sorted_list = wp_list_sort( $url_list, 'lastmod', 'DESC' );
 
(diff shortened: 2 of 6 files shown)


inc/class-core-sitemaps-provider.php
```diff
@@ -194,15 +216,136 @@ public function get_sitemaps() {
...
+	/**
+	 * Calculate lastmod date for a sitemap page.
+	 *
+	 * Calculated value is saved to the database as an option.
+	 *
+	 * @param string $type    The object type of the page: posts, taxonomies, users, etc.
+	 * @param string $subtype The object subtype if applicable, e.g., post type, taxonomy type.
+	 * @param int    $page    The page number.
+	 */
+	public function calculate_sitemap_lastmod( $type, $subtype, $page ) {
```
joemcgill, line 329: I've opened #91 to address confusing naming of parameters and properties that I'd like to see addressed, but is outside of the scope of this issue.

inc/class-core-sitemaps-provider.php
```diff
@@ -194,15 +216,133 @@ public function get_sitemaps() {
...
+
+			// Handle lists of post-objects.
+			if ( isset( $type->name ) ) {
+				$name = $type->name;
+			}
+
+			$total = $this->max_num_pages( $name );
+
+			for ( $page = 1; $page <= $total; $page ++ ) {
+				wp_schedule_single_event( time(), 'core_sitemaps_calculate_lastmod', array( $this->slug, $name, $page ) );
```
joemcgill: We could check for an already scheduled event here as well.

inc/class-core-sitemaps-provider.php
```diff
@@ -42,6 +42,23 @@ class Core_Sitemaps_Provider {
 	 */
 	public $slug = '';
 
+	/**
+	 * Set up relevant rewrite rules, actions, and filters.
+	 */
+	public function setup() {
```
joemcgill, line 48: I've moved the setup for all actions performed by providers from the main sitemap class and into the provider itself to keep responsibilities clear.

inc/class-core-sitemaps-provider.php
```diff
@@ -194,15 +196,63 @@ public function get_sitemaps() {
...
 	}
 
+	/**
+	 * Get the URL of a sitemap entry.
+	 *
+	 * @param string $name The name of the sitemap.
+	 * @param int    $page The page of the sitemap.
+	 * @return string The composed URL for a sitemap entry.
+	 */
+	public function get_sitemap_url( $name, $page ) {
```
svandragt, line 267: nice encapsulation!

inc/class-core-sitemaps-index.php
```diff
@@ -70,8 +70,16 @@ public function render_sitemap() {
 		$sitemap_index = get_query_var( 'sitemap' );
 
 		if ( 'index' === $sitemap_index ) {
-			$sitemaps = core_sitemaps_get_sitemaps();
-			$this->renderer->render_index( array_keys( $sitemaps ) );
+			$providers = core_sitemaps_get_sitemaps();
+
+			$sitemaps = array();
+
+			// Build up the list of sitemap pages.
```
svandragt: Don't think the comment is required, the code is very clear.

inc/class-core-sitemaps-provider.php
```diff
@@ -245,12 +256,52 @@ public function get_sitemap_url( $name, $page ) {
...
+
+		$times = wp_list_pluck( $list, 'lastmod' );
+
+		usort( $times, function( $a, $b ) {
+			return strtotime( $b ) - strtotime( $a );
+		} );
+
+		$suffix = implode( '_', array_filter( array( $type, $subtype, (string) $page ) ) );
+
+		update_option( "core_sitemaps_lasmod_$suffix", $times[0] );
```
svandragt: This looks like an intentional misspelling of lastmod, it might be worth commenting on that as it caught me by surprise thinking it was a type; or revising the name of the option.

inc/class-core-sitemaps-provider.php
```diff
@@ -245,12 +256,52 @@ public function get_sitemap_url( $name, $page ) {
...
-		return '0000-00-00 00:00Z GMT';
+		$type = implode( '_', array_filter( array( $this->slug, $name, (string) $page ) ) );
+
+		// Check for an option.
+		$lastmod = get_option( "core_sitemaps_lasmod_$type", '' );
+
+		// If blank, schedule a job.
+		if ( empty( $lastmod ) && ! wp_doing_cron() ) {
+			wp_schedule_single_event( time() + 500, 'core_sitemaps_calculate_lastmod', array( $this->slug, $name, $page ) );
+		}
```
svandragt, line 315: This can result in a blank lastmod value rendering into the sitemap. Consider setting a fallback value to the current date at midnight in W3c datetime, after scheduling a job, so that there is always a valid timestamp.

inc/class-core-sitemaps-provider.php
```diff
@@ -295,9 +296,12 @@ public function calculate_sitemap_lastmod( $type, $subtype, $page ) {
 
 		$times = wp_list_pluck( $list, 'lastmod' );
 
-		usort( $times, function( $a, $b ) {
-			return strtotime( $b ) - strtotime( $a );
-		} );
+		usort(
+			$times,
+			function( $a, $b ) {
```
svandragt: This closure can be declared as static to improve preformance.

inc/class-core-sitemaps-provider.php
```diff
@@ -308,6 +316,31 @@ function( $a, $b ) {
...
+			$name = $type;
+
+			// Handle lists of post-objects.
+			if ( isset( $type->name ) ) {
+				$name = $type->name;
+			}
+
+			$total = $this->max_num_pages( $name );
+
+			for ( $page = 1; $page <= $total; $page ++ ) {
```
svandragt: FYI (outside the scope of this PR) There is protential here to reduce code duplication by passing in a callback into an abstracted function that calls the callback for each page, as this code is identical to `get_sitemap_entries()`.

inc/class-core-sitemaps-provider.php
```diff
@@ -279,7 +279,12 @@ public function get_sitemap_lastmod( $name, $page ) {
 
 		// If blank, schedule a job.
 		if ( empty( $lastmod ) && ! wp_doing_cron() ) {
-			wp_schedule_single_event( time() + 500, 'core_sitemaps_calculate_lastmod', array( $this->slug, $name, $page ) );
+			$event_args = array( $this->slug, $name, $page );
+
+			// Don't schedule a duplicate job.
+			if ( ! wp_next_scheduled( 'core_sitemaps_calculate_lastmod', $event_args ) ) {
```
svandragt, line 312: It's good to be extra safe, although if the codex (https://codex.wordpress.org/Function_Reference/wp_schedule_single_event) is to be believed it's currently not needed:

>Note that scheduling an event to occur before 10 minutes after an existing event of the same name will be ignored, unless you pass unique values for $args to each scheduled event. See wp_next_scheduled() for more information.
>
>This behavior is subject to change, as the original intention of the code was to prevent scheduling two identical events within ten minutes of each other, not preventing scheduling of identical events until ten minutes after the next scheduled occurrence.

inc/class-core-sitemaps-provider.php
```diff
@@ -283,7 +283,7 @@ public function get_sitemap_lastmod( $name, $page ) {
 
 			// Don't schedule a duplicate job.
 			if ( ! wp_next_scheduled( 'core_sitemaps_calculate_lastmod', $event_args ) ) {
-				wp_schedule_single_event( time() + 500, 'core_sitemaps_calculate_lastmod', $event_args );
+				wp_schedule_single_event( time(), 'core_sitemaps_calculate_lastmod', $event_args );
```
svandragt, line 313: The `+ 500` prevents running multiple events within the same 500s time period as an event is scheduled to run, by removing this the code is relying on the behaviour described in the codex that is subject to change.

inc/class-core-sitemaps-provider.php
svandragt: See above about the 500s delay.

inc/class-core-sitemaps-provider.php
```diff
@@ -42,6 +42,23 @@ class Core_Sitemaps_Provider {
...
+		// Set up rewrite rules and rendering callback.
+		add_rewrite_rule( $this->route, $this->rewrite_query(), 'top' );
+		add_action( 'template_redirect', array( $this, 'render_sitemap' ) );
+
+		// Set up async tasks related to calculating lastmod data.
+		add_action( 'core_sitemaps_calculate_lastmod', array( $this, 'calculate_sitemap_lastmod' ), 10, 3 );
+		add_action( 'core_sitemaps_update_lastmod_' . $this->slug, array( $this, 'update_lastmod_values' ) );
+
+		if ( ! wp_next_scheduled( 'core_sitemaps_update_lastmod_' . $this->slug ) && ! wp_installing() ) {
+			wp_schedule_event( time(), 'twicedaily', 'core_sitemaps_update_lastmod_' . $this->slug );
```
svandragt: Consider making the schedule `twicedaily` filterable, as depending on the use case spidering twice a day is not enough, eg news. (assuming the sitemap's last modified date influences whether the sitemap page should be explored by the crawler).

svandragt (changes requested):
Hey Joe, I've been a bit strict in this review because I thought it would be worth erring on the side of caution here; so feel free to consider my suggestions and choose to not implement them.

On the whole I believe the design is sound and will be performant on shared hosting sites with a small number of sitemaps using wp-cron, as well as larger sites with a task scheduler like cavalcade to perform async updates.

Excellent work.

inc/class-core-sitemaps-provider.php
joemcgill: This is the intended behavior. Since `lastmod` values for sitemap pages is an expensive operation, we want to show no value when we don't have a previously calculated value, so we aren't blocking rendering of the page. When this is the case, we calculate this value asynchronously so it's available the next time the index is displayed.

inc/class-core-sitemaps-provider.php
joemcgill: Good catch! this is _not_ intentional, I just misspelled in 🤦‍♂

inc/class-core-sitemaps-provider.php
joemcgill: Yeah, I thought of that as well.

inc/class-core-sitemaps-provider.php
joemcgill: `wp_schedule_single_event()` indeed takes care of making sure duplicates are not scheduled within 10 minutes +/- one that is already scheduled. In this case, I'm trying to avoid scheduling a duplicate job _ever_. Perhaps unnecessary, but that's why I've added this check.

inc/class-core-sitemaps-provider.php
joemcgill: I don't think this is accurate. All that `+ 500` does is change the timestamp from "now" to 500 seconds later than "now". It won't have any effect on running multiple events beyond what `wp_schedule_single_event()` does internally to avoid duplicates within 10 minutes. I don't see a need to delay the event by 500 seconds, which is why I removed that value (which I initially added to give me a buffer to validated that all the events were actually scheduled).

inc/class-core-sitemaps-provider.php
joemcgill: I've simplified this using `wp_list_sort()` in e9b24b4.

joemcgill:
@svandragt this is ready for another review when you have a chance. Thanks for all the feedback!

inc/class-core-sitemaps-provider.php
svandragt: Ah I missed that `lastmod` is an optional element, so it can be omitted safely. All good.

inc/class-core-sitemaps-provider.php
svandragt: Ok, happy with you having looked into it.

svandragt:
This is such a nice project and the PR is a really good improvement 👍 Nice job @joemcgill

svandragt (approved):

PR #102, opened 2020-01-10
Fix gathering sitemap entries before rendering the index.

joemcgill:
### Description
This fixes a bug introduced in d1a826c, which caused sitemap entries to be an empty list before sending them to the renderer in `Core_Sitemaps_Index ::render_sitemap()`.

### Type of change

### Steps to test
Before applying this branch the main XML file should not include any sitemap entries. Applying this branch causes them all to be displayed again.

### Acceptance criteria

### DIFF
--- a/inc/class-core-sitemaps-index.php
+++ b/inc/class-core-sitemaps-index.php
@@ -76,7 +76,7 @@ class Core_Sitemaps_Index {
 
 			foreach ( $providers as $provider ) {
 				// Using array_push is more efficient than array_merge in a loop.
-				$sitemaps = array_push( $sitemaps, ...$provider->get_sitemap_entries() );
+				array_push( $sitemaps, ...$provider->get_sitemap_entries() );
 			}
 
 			$this->renderer->render_index( $sitemaps );


joemcgill:
@swissspidy I'm wondering about abstracting this functionality, which is also used in a few unit tests into a helper function like:

```
function core_sitemaps_get_sitemap_entries() {
	$sitemaps  = array();
	$providers = core_sitemaps_get_sitemaps();

	foreach ( $providers as $provider ) {
		// Using array_push is more efficient than array_merge in a loop.
		array_push( $sitemaps, ...$provider->get_sitemap_entries() );
	}

	return apply_filters( 'filter_name', $sitemaps );
}
```

I had originally avoided doing so because I want to limit how many public functions are made available for back-compat reasons, and I don't know of a use case for using this outside of the specific context where we're rendering the sitemap. 

However, abstracting this into a function would allow us to make that functionality testable as a unit and would have guarded against this bug.

Thoughts?

swissspidy (approved):

swissspidy:
Hmm I am a bit on the fence on that one. If deemed valuable we can always add such a function at a later stage. But for now I think we can do without. Let's track it in a new issue though 👍

joemcgill:
Thanks @swissspidy. I've opened #103 to document the idea of abstracting the functionality for gathering sitemap entries for the index and will merge/close this.

PR #105, opened 2020-01-20
Consolidate rewrites and rendering

joemcgill:
### Issue Number
Blocked by #87. Once that is merged, this should be updated to point to 'master'.

### Description
This consolidates redundant functionality for registering rewrites and rendering files from individual providers and makes them a concern of the main Core_Sitemaps class.

Changes:

- Adds a new `renderer` property to `Core_Sitemaps`, which references a `Core_Sitemaps_Renderer` object.
- Removes references to `Core_Sitemaps_Renderer` objects from the `Core_Sitemaps_Index` and `Core_Sitemaps_Provider` objects.
- Moves all registration of rewrite tags and routes to a new `Core_Sitemaps::register_rewrites()` method.
- Consolidates all `template_redirect` rendering callbacks to a single `Core_Sitemaps::render_sitemaps()`.
- Renames `Core_Sitemaps::xsl_stylesheet_rewrites()` to `Core_Sitemaps::register_xsl_rewrites()` to be more declarative and match the new `Core_Sitemaps::register_rewrites()` method.

### Type of change
Please select the relevant options:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Enhancement (change which improves an existing feature. E.g., performance improvement, docs update, etc.)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Steps to test
All unit tests are still passing and sitemap pages are being rendered as expected.

### Acceptance criteria
- [x] My code follows WordPress coding standards.
- [x] I have performed a self-review of my own code.
- [ ] If the changes are visual, I have cross browser / device tested.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [x] I have added test instructions that prove my fix is effective or that my feature works.

PR #104, opened 2020-01-17
Merge pull request #104 from GoogleChromeLabs/feature/87-add-sitemaps

joemcgill:
This allows for adding new sitemaps to the system using a new `core_sitemaps_register_sitemap()` helper function.

Added:

- `core_sitemaps_register_sitemap()` – Used to register a new sitemap provider.
- Hook `core_sitemaps_register` – Fires after core sitemaps are registered so new sitemaps can be added or additional actions can be taken.
- `Core_Sitemaps_Tests::test_register_sitemap_provider` – test method to confirm registered sitemaps are included in the registry.
- `Core_Sitemaps_Test_Provider` class used in the PHPUnit test suite.

### Issue Number
Closes #87.

### Type of change
Please select the relevant options:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Enhancement (change which improves an existing feature. E.g., performance improvement, docs update, etc.)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Steps to test
Add a new callback on the `core_sitemaps_register` hook that registers a new sitemap provider and see that it shows up in the list of sitemaps. For example:

```
add_action( 'core_sitemaps_register', function () {
	core_sitemaps_register_sitemap( 'test', new Core_Sitemaps_Provider() );

	var_dump( core_sitemaps_get_sitemaps() );
	die();

} );
```

You should see that a sitemap named 'test' is included in the list of providers.

### Acceptance criteria
- [x] My code follows WordPress coding standards.
- [x] I have performed a self-review of my own code.
- [ ] If the changes are visual, I have cross browser / device tested.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [x] I have added test instructions that prove my fix is effective or that my feature works.

PR #106, opened 2020-01-20
Initialize Core Sitemaps system on the init hook.

joemcgill:
### Issue Number
Blocked by #104 and will need to be updated once #105 is merged.

### Description
This creates a new helper function for getting the global `Core_Sitemaps` instance and uses this function to initialize the whole system via the WordPress `init` hook.

Changes:

- Adds `core_sitemaps_get_server()`, which returns the global instance of `Core_Sitemaps` system and initialize one if none already exists.
- Adds `core_sitemaps_init` action that fires after the global `Core_Sitemaps` object is initialized.
- Renames `Core_Sitemaps::bootstrap()` to `Core_Sitemaps::init()` to reinforce that it fires on the init hook.
- Updates the firing/registration of callbacks that formerly were registered in `Core_Sitemaps::bootstrap()`
- Updates all helper functions that were referencing the global $core_sitemaps object to use `core_sitemaps_get_server()` instead.

### Type of change

### Acceptance criteria

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -6,12 +6,45 @@
  */
 
 /**
+ * Retrieves the current Sitemaps server instance.
+ *
+ * @return Core_Sitemaps Core_Sitemaps instance.
+ */
+function core_sitemaps_get_server() {
+	/**
+	 * Global Core Sitemaps instance.
+	 *
+	 * @var Core_Sitemaps $core_sitemaps
+	 */
+	global $core_sitemaps;
+
+	// If there isn't a global instance, set and bootstrap the sitemaps system.
+	if ( empty( $core_sitemaps ) ) {
+		$core_sitemaps = new Core_Sitemaps();
+		$core_sitemaps->init();
+
+		/**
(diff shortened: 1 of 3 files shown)


inc/functions.php
```diff
@@ -5,13 +5,46 @@
...
+		/**
+		 * Fires when initializing the Core_Sitemaps object.
+		 *
+		 * Additional sitemaps should be registered on this hook.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param core_sitemaps $core_sitemaps Server object.
+		 */
+		do_action( 'core_sitemaps_init', $core_sitemaps );
```
swissspidy, line 35: Just so it's clear, this is now the replacement for `core_sitemaps_register`?

joemcgill: Correct. I think this action pattern is more clear and allows functionality to be extended/modified when the system has initialized, and can be used internally by the sitemaps system to better control the flow of actions that are initialized specifically by the sitemaps system.

swissspidy (approved):

PR #117, opened 2020-02-02
Fix esc_html with translation

szepeviktor:
Discovered by @phpstan

### Description
Fix 2 esc_html calls with translation

### Type of change


swissspidy:
Why this change for exactly these 2 cases?

Note that WordPress core doesn‘t escape translations like this and as a feature plugin this project shouldn‘t either.

szepeviktor:
There are **two parameters.** esc_html does not handle translations [esc_html__](https://developer.wordpress.org/reference/functions/esc_html__/) does. These were typos.

swissspidy:
Gotcha! Was still early I guess 😅

swissspidy (approved):

szepeviktor:
👍

PR #119, opened 2020-02-02
Introduce PHPStan to CI (tests should fail!)

szepeviktor:
### Description
add static analysis to CI

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Enhancement (change which improves an existing feature. E.g., performance improvement, docs update, etc.)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #121, opened 2020-02-02
WP_Query::get_posts may return int[]|WP_Post[]

szepeviktor:
### Description
not possible to statically determine - is up to `fields` query argument

### Type of change


swissspidy:
Since we don't use `fields`, shouldn't it be safe to assume that those are post objects only? Otherwise there would be errors further down in the foreach loop and we should add some safeguards

szepeviktor:
> safe to assume that those are post objects only?

For human, yes.
For static analysis, no -> #119

swissspidy (approved):

Issue #128, opened 2020-02-19
prefix sitemap(-*).xml to wp-sitemap(-*).xml


pfefferle:
### Description

To prevent collisions with existing plugins, we discussed to rename the `sitemap(-*).xml` to `wp-sitemap(-*).xml`.

The file-name(s) should be also filterable, to help other plugins, that use the new core feature, to be consistent.

swissspidy:
I am not sure whether we should make that prefix filterable. We could just as well hardcode it.

With what would this be consistent?

pfefferle:
If you use one of the current plugins, they mostly generate a `/sitemap.xml` and this might cause some 404s and that could cause issues with Search engines (even if the redirect is implemented). Why not make it filterable? What is the downside?

swissspidy:
https://github.com/GoogleChromeLabs/wp-sitemaps/pull/129#issuecomment-590915627 suggests redirecting requests to `/sitemap.xml` if they'd lead to 404s. So that would be covered.

> Why not make it filterable? What is the downside?

Unnecessary overhead and error-proneness for no real benefit. The sitemap name does not matter for consumers. Worst case there's a redirect or there are two sitemaps running on the site. Still, the correct sitemap would be referenced in robots.txt

pfefferle:
My intention was: Plugin developer like [Google Sitemaps](https://wordpress.org/plugins/google-sitemap-generator/) will be able to keep the existing `/sitemap.xml` slug if they use/extend the core implementation. But true, if we implement https://github.com/GoogleChromeLabs/wp-sitemaps/pull/129#issuecomment-590915627 it shouldn't be a problem.

PR #129, opened 2020-02-21
Use wp-sitemap(.*).xml instead of sitemap(.*).xml

pfefferle:
### Issue Number
#128 

### Description
Use `wp-sitemap(.*).xml` instead of `sitemap(.*).xml` and add a possibility to filter the filename prefix.

### Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Acceptance criteria


inc/functions.php
```diff
@@ -49,6 +49,20 @@ function core_sitemaps_get_sitemaps() {
 	return $core_sitemaps->registry->get_sitemaps();
 }
 
+/**
+ * Retrieves the prefix for the sitemap filename.
+ *
+ * @return string Prefix.
+ */
+function core_sitemaps_filename_prefix() {
```
swissspidy: `filename` might be misleading because they're not actually files. Maybe just `core_sitemaps_prefix`? Or perhaps `core_sitemaps_sitemap_prefix`

pfefferle: `core_sitemaps` will be removed when merging into core!?! If so, I would prefer `core_sitemaps_sitemap_prefix`.

swissspidy: That hasn't been decided yet. We might also just replace `core_sitemaps` with `wp_sitemaps` or end up using something like `wp_sitemap_prefix` in the end. But this doesn't matter for the sake of adding it to the plugin 🙂

kraftbj:
I'd suggest also hooking into pre_handle_404 while a feature plugin and patch `handle_404` during Core merge. If someone attempts /sitemap.xml (or other supported formats) and it would 404, let's redirect them to the wp-sitemap.xml.

That would help discoverability of the sitemap ("sitemap.xml" being pretty common, like "feed" is for RSS) while not conflicting with existing solutions.

I'm happy to work on a patch for this to be applied to this branch (or master if post merge).

swissspidy:
Added an alternative PR without the filter here: #135

pfefferle:
closed in favor of #135

PR #136, opened 2020-03-03
Add plugin activation and deactivation hooks

### LINKED ISSUE #132: Flush rewrite rules upon plugin (de)activation
As mentioned on Slack, we might need to flush rewrite rules using `flush_rewrite_rules( false )` upon plugin activation and deactivation to prevent potential issues with rewrites and other plugins.

Related: #128

swissspidy:
### Issue Number

Fixes #132.

### Description

Add plugin activation and deactivation hooks to flush rewrites appropriately, to minimize conflicts with other plugins.

### Type of change

### Acceptance criteria

### DIFF
--- a/core-sitemaps.php
+++ b/core-sitemaps.php
@@ -47,3 +47,31 @@ require_once __DIR__ . '/inc/functions.php';
 
 // Boot the sitemaps system.
 add_action( 'init', 'core_sitemaps_get_server' );
+
+/**
+ * Plugin activation hook.
+ *
+ * Adds and flushes rewrite rules.
+ */
+function core_sitemaps_plugin_activation() {
+	$core_sitemaps = new Core_Sitemaps();
+	$core_sitemaps->register_rewrites();
+	$core_sitemaps->register_xsl_rewrites();
+	flush_rewrite_rules( false );
+}
+
+register_activation_hook( __FILE__, 'core_sitemaps_plugin_activation' );
+
+/**
+ * Plugin deactivation hook.
+ *
+ * Adds and flushes rewrite rules.
+ */
+function core_sitemaps_plugin_deactivation() {
+	$core_sitemaps = new Core_Sitemaps();
+	$core_sitemaps->register_rewrites();
+	$core_sitemaps->register_xsl_rewrites();
+	flush_rewrite_rules( false );
+}
+
+register_deactivation_hook( __FILE__, 'core_sitemaps_plugin_deactivation' );


pfefferle (approved):

pbiron:
I was coming to comment on this...and I noticed you just merged it a few minutes ago :-(

The activation hook works as expected, but the deactivation  hook does not.

You can see this by installing/activating the [Rewrite Rules Inspector](https://wordpress.org/plugins/rewrite-rules-inspector/) plugin.  With the sitemaps plugin active, go to `Tools > Rewrite Rules`, enter `https://example.com/sitemap.xml` (where `example.com` is the domain for your test site) in the "Match URL" field and click `Filter`.  You'll see the rule for `^wp-sitemap\.xml$`.  Then, deactivate the sitemaps plugin and click the `Filter` button again in `Tools > Rewrite Rules`.  You'll still see the rule for `^wp-sitemap\.xml$`.

First, there is no need to call `$core_sitemaps->register_rewrites();` and `$core_sitemaps->register_xsl_rewrites();`...they've already been called by the time the deactivation hook is fired.  But even if they hadn't already been called, "registering" the plugin's rewrite rules is **not** what should happen in the deactivation handler.

Second, the rewrite rules that have been added by the plugin need to be "unregistered" **before** the call to `flush_rewrite_rules()`.

Unfortunately, there is no core function/method that is the opposite of `add_rewrite_rule()` (there is a `remove_rewrite_tag()` tho) .  The only way I've found to do it is to call `unset( $wp_rewrite->extra_rules_top['^wp-sitemap\.xml$'] );`. etc.

Therefore, what I suggest is adding a couple of methods to the `Core_Sitemaps` class: `unregister_rewrites()` and `unregister_xsl_rewrites()`.  In those 2 methods, do the appropriate `unset()`s and then call those methods in the deactivation handler before the call to `flush_rewrite_rules()`.

I can do a PR for that, but wanted to have someone check my logic above first.

pfefferle:
You are right, I +1 too fast. The `core_sitemaps_plugin_deactivation` has to simply call `flush_rewrite_rules( false );`.

pfefferle:
When the `deactivation_hook` is called, the plugin is no longer loaded, so the rewrite rules are no longer loaded and the `flush_rewrite_rules( false );` should flush them.

pfefferle:
`register_deactivation_hook( __FILE__, 'flush_rewrite_rules' );` should also work!?!

swissspidy:
No idea what I was thinking 🤦‍♂ Thanks for noticing @pbiron!

@pfefferle Are you sure about that? See https://core.trac.wordpress.org/ticket/29118

swissspidy:
> `register_deactivation_hook( __FILE__, 'flush_rewrite_rules' );` should also work

No, that performs a hard flush, which we wouldn't want

pbiron:
> When the `deactivation_hook` is called, the plugin is no longer loaded, so the rewrite rules are no longer loaded and the `flush_rewrite_rules( false );` should flush them.

Try the `Rewrite Rules Inspector` as explained above and you'll see that it doesn't work.

swissspidy:
Let me push another PR for this 🙂

pbiron:
See this answer from Otto to an old WP.stackexchange question : https://wordpress.stackexchange.com/a/104106/113496, which is where I learned to `unset()` rewrite rules on deactivate before flushing.

pfefferle:
> @pfefferle Are you sure about that? See https://core.trac.wordpress.org/ticket/29118

Hmmm... Then I have to check a lot of my plugins... :(

pbiron:
The reason is that `WP_Rewrite::flush_rules()` does a `$this->extra_rules_top = array_merge( $this->extra_rules_top, $rules );` and by the time the deactivation hook is called `$wp_rewrite->extra_rules_top` still has the plugin's rewrite rules.

pbiron:
The other thing is: with the activation hook, `Core_Sitemaps::init()` does **not** need to hook into `wp_loaded` and the `maybe_flush_rewrites()` method can be removed.

This is because updating the plugin when it is activate will 1) deactivate it before the update; 2) activate it after the update.  So, checking the version number to decide whether the flush rewrites is no longer necessary.

swissspidy:
@pbiron Now that we changed the URLs from `sitemap.xml` to `wp-sitemap.xml`, the `maybe_flush_rewrites` function is very useful

PR #137, opened 2020-03-03
Make query instances filterable

### LINKED ISSUE #131: Add filter to modify WP_Query arguments
### Description

Similar to what's suggested in https://wordpress.slack.com/archives/CTKTGNJJW/p1582058607135400, it might be easier to include/exclude to be able to adjust the `WP_Query` instance instead of having to rely on the `core_sitemaps_*_url_list` filters.

If you want to exclude a lot of posts from the sitemap, it's better to not query the DB for them in the first place.

`do_action( 'core_sitemaps_...', &$query)` everywhere we use  `WP_Query` and `WP_Term_Query`.

**Additional context**

### Acceptance Criteria
- []
- []
- []

swissspidy:
### Issue Number
Fixes #131.

### Description

Allows developers to modify queries to exclude posts/users/terms and things like that.

### Type of change

### Acceptance criteria


pfefferle:
Shouldn‘t this be filters instead of actions?

swissspidy:
Not really, as long as the vars are passed by reference. Which I didn't do 🤦‍♂ 

See this for comparison:

https://github.com/WordPress/wordpress-develop/blob/60f78cd845a4725b0ba0c349666362c7923735c0/src/wp-includes/class-wp-query.php#L1761-L1772

pfefferle:
I prepared a MR for the "by reference" changes... Should I commit them to your branch?

swissspidy:
@pfefferle Sure, feel free! Happy to take a look

pfefferle:
@googlebot I consent

joemcgill:
Let's make sure to add some test cases along with this.

pbiron:
Just starting to review this and evaluate how well the new actions support a couple of use cases I have.

I've started with the `core_sitemaps_user_query` action.  I mentioned the following use case in slack during the 1st meeting, but I'll repeat it here for completeness:

1. I've got a site with a large number of users who are "members" in an organization, the vast majority of whom don't have publish posts
2. The site has an "Our Members" page and sets `author_base` to the slug of that page (so that `get_author_posts_url()` returns URLs such as `https://example.com/our-members/jane-doe/`)
3. The content of the "Our Members" page is a custom block that displays links to member "profiles", categorized in various ways (e.g., by name, by medium [the members are artists, so the mediums are "Oil", "Pastel", "Sculpture", "Photography", etc])
4. The member profiles are generated from info stored in user meta
5. I would like the users sitemap to show not users who have published posts, but those who have a "published" profile (not all members have have entered the requisite info for their profile, those with a "published" profile are also determined by the presense of a specific user meta)

I can currently support this use case with the exsting `core_sitemaps_users_url_list` filter (to completely replace the URLs generated by the users provider) and/or by registering a custom provider (with `core_sitemaps_register_providers`...and unregistering the users provider), but I was hoping that these new actions would make it easier (and/or more efficient) than either of those 2 methods.

Here are a couple observations on this new action:

1. by the time the new `core_sitemaps_user_query` action is fired, the unnecessary (in my case) 'has_published_posts' database query has already been executed (which is also why using `core_sitemaps_users_url_list` is less than ideal)
2. this could be avoided by changing to something like
```
$query = new WP_User_Query();
$query->prepare_query(
	array(
		'has_published_posts' => array_keys( $public_post_types ),
		'number'              => core_sitemaps_get_max_urls( $this->slug ),
		'paged'               => absint( $page_num ),
	)
);
do_action_ref_array( 'core_sitemaps_user_query', array( &$query ) );
$query->query();
```
3. but even then, the callback I hook to `core_sitemaps_user_query` has to call `$query->prepare_query()` again after modifying the `query_vars`.  I suspect that many developers are not going to realize they need to do that and will wonder why the changes they make to `query_vars` in their callback don't have any effect

The above observations are why I'd much prefer the use of filters for these kind of query modifications.  That is:

```
$args = array(
	'has_published_posts' => array_keys( $public_post_types ),
	'number'              => core_sitemaps_get_max_urls( $this->slug ),
	'paged'               => absint( $page_num ),
);
$args  = apply_filters( 'core_sitemaps_user_query_args', $arg );
$query = new WP_User_Query( $args );
```

seems much more straight forward.  As a developer, it's easier for me to think in terms of modifying the arguments to the query before it is run than to modify the query object after it has run.

Of course, there may be other use cases where an action like this is preferable, so there is no reason we can't include both: filter the args before the query is run **and** modify the query object after the query is run.

I'll continue to explore how these new actions impact other use cases I have and will likely have more comments.

joemcgill:
Thanks for the research @pbiron. I agree with you that filtering the arguments are much more straightforward than filtering the `WP_Query` object. Another pattern we could also try is to have a filter that runs before the query that evaluates to `null` or `false` by default, which allows people to swap out their own querying logic entirely—bypassing the default query. This would allow for much more control if people needed to write custom queries rather than relying on a complex `WP_Query` instance.

pfefferle:
I would prefer the last option (to filter the query array).

swissspidy:
After reading through last week's chat, I am closing this PR in favor of a new PR that adds filters for query arguments as well as pre-query filters.

PR #138, opened 2020-03-03
Bail early if search engines are discouraged

### LINKED ISSUE #111: Disable sitemap generation if site is private.
### Description

Currently the site map reference is not added to `robots.txt` if `! get_option( 'blog_public' )`. In such cases, it may be better to bypass sitemap generation in its entirety.

Pros:
* Sitemap will not instruct search engines to visit pages with `noindex` meta tags
* Less work for the server

Cons:
* Upon setting the blog to public, the sitemap will not exist

#24 

Adding this to the bootstrapping process would do the tick. The `do_robots` action would need to be removed too.

```php
if ( ! get_option( 'blog_public' ) {
	return;
}
```

**Additional context**

### Acceptance Criteria

- [ ] Private sites do not have sitemaps
- [ ] Sitemap is not listed in robots.txt for private sites

### LINKED ISSUE #123: Filter to disable XML Sitemaps feature
If there is not one make sure there is a filter that can be used to disable the XML sitemaps feature.

swissspidy:
### Issue Number
Fixes #111.
Fixes #123.

### Description

Does not do anything for private sites.

### Type of change

### Test Instructions

1. Activate plugin
1. Make site private
1. Verify that sitemap is not being generated

### Acceptance criteria


phpcs.xml.dist
```diff
@@ -31,7 +31,10 @@
 	<!-- Rules: WordPress Coding Standards -->
 	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards -->
 	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties -->
-	<rule ref="WordPress"/>
+	<rule ref="WordPress-Core"/>
+	<rule ref="WordPress-Docs">
+        <exclude-pattern>tests/*</exclude-pattern>
+    </rule>
```
pfefferle: wrong indents

tests/phpunit/class-test-core-sitemaps.php
```diff
@@ -334,7 +334,23 @@ public function test_robots_text() {
...
+	}
+
+	/**
+	 * Test robots.txt output for a private site.
+	 */
+	public function test_robots_text_private_site() {
+		// Simulate private site (search engines discouraged).
+		update_option( 'blog_public', '0' );
+
+		$robots_text = apply_filters( 'robots_txt', '', true );
```
pfefferle: shouldn't the last param be false?

pfefferle: the site is not public in this case

swissspidy: Right! Copy paste strikes again...

Although it shouldn't make a difference here, as our code should bail way earlier, i.e. `robots_txt` should never be filtered if the site is not public.

pfefferle: Maybe a race condition? The filter is set by the test before and `update_option( 'blog_public', '0' );` does not unset the function from the filter list?!?

swissspidy: Yeah the plugin is loaded before all tests run, so this is to be expected

pfefferle: But then `update_option( 'blog_public', '0' );` should be removed (because it is not really tested here) or the plugin has to be "reloaded" here, to test the option flag.

swissspidy: Yep, I still need to get to that 👍

pfefferle: hmmm... but if we want to test `update_option( 'blog_public', '0' );` we have to use `$robots_text = apply_filters( 'robots_txt', '', true );`... otherwise it does not test the option.

So we need:

```PHP
	/**
	 * Test robots.txt output for a private site.
	 */
	public function test_robots_text_private_site() {
                // @todo rest plugin

		// Simulate private site (search engines discouraged).
		update_option( 'blog_public', '0' );

		$robots_text = apply_filters( 'robots_txt', '', true );
		$sitemap_string = 'Sitemap: http://' . WP_TESTS_DOMAIN . '/?sitemap=index';

		// Simulate public site.
		update_option( 'blog_public', '1' );

		$this->assertNotContains( $sitemap_string, $robots_text );
	}
```

and

```PHP
	/**
	 * Test robots.txt output for a private site.
	 */
	public function test_robots_text_private_site() {
		$robots_text = apply_filters( 'robots_txt', '', false );
		$sitemap_string = 'Sitemap: http://' . WP_TESTS_DOMAIN . '/?sitemap=index';

		$this->assertNotContains( $sitemap_string, $robots_text );
	}
```

inc/functions.php
```diff
@@ -26,20 +26,34 @@ function core_sitemaps_get_server() {
...
-		do_action( 'core_sitemaps_init', $core_sitemaps );
+	/**
+	 * Filters whether XML Sitemaps are enabled or not.
+	 *
+	 * @param bool $is_enabled Whether XML Sitemaps are enabled or not. Defaults to true for public sites.
+	 */
+	$is_enabled = (bool) apply_filters( 'core_sitemaps_is_enabled', $is_enabled );
+
+	if ( ! $is_enabled ) {
+		return $core_sitemaps;
```
swissspidy: Looks like this breaks everything 😅

joemcgill: I think it would be better to put this check early, before the `Core_Sitemaps` object is instantiated, and return either null or `false` if the server is disabled.

svandragt:
The theoretical concern I have is that you might want to submit a sitemap before the site is launched because indexing can take a while (to get a good search presence in the launch window), and this PR prevents that possibility.

Now I'm not sure whether the search engine being aware of the sitemap but not able to crawl any of the pages due to the site being discouraged results in faster / more complete recrawl when the site is available for indexing, but I would think it might.

Generally, could we detail the use case for this feature? Having signposts for a shop that's currently not open isn't a problem.

joemcgill:
@svandragt the current approach allows `core_sitemaps_is_enabled` to be filtered. Does this cover your concern?

svandragt:
@joemcgill Yep looks good to me.

pbiron (approved):
LGTM.

PR #139, opened 2020-03-03
Fix plugin deactivation

swissspidy:
### Issue Number

See #132, #136

### Description

See https://github.com/GoogleChromeLabs/wp-sitemaps/pull/136#issuecomment-594128540 for details. Props @pbiron

### DIFF
--- a/core-sitemaps.php
+++ b/core-sitemaps.php
@@ -56,7 +56,6 @@ add_action( 'init', 'core_sitemaps_get_server' );
 function core_sitemaps_plugin_activation() {
 	$core_sitemaps = new Core_Sitemaps();
 	$core_sitemaps->register_rewrites();
-	$core_sitemaps->register_xsl_rewrites();
 	flush_rewrite_rules( false );
 }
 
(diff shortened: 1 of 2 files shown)


pbiron:
+1 on folding `register_xsl_rewrites()` into `register_rewrites()`.

Just testing that the new PR works as expected.

pbiron:
I think the `unregister_rewrites()` method should also call:

```
remove_rewrite_tag( '%sitemap%' );
remove_rewrite_tag( '%sub_type%' );
remove_rewrite_tag( '%stylesheet%' );
```

Other than that looks good to me.

swissspidy:
@pbiron Rewrite tags are not persisted in the DB though, are they? So that seemed a bit unnecessary

pbiron:
> @pbiron Rewrite tags are not persisted in the DB though, are they? So that seemed a bit unnecessary

I'm not sure...there is a prop for them in `WP_Rewrite`...but that may just be per-invocation and not persisted.

swissspidy:
Rewrite tags are basically just placeholders for regex patterns that get replace when generating the rules. Only the rules generated by `\WP_Rewrite::rewrite_rules()` are then stored. So this seems correct to me.

pfefferle (approved):
thanks @pbiron for testing!

swissspidy:
Thanks to both of you!

PR #142, opened 2020-03-05
Use wp_die when SimpleXML is not available

### LINKED ISSUE #122: Clarify the SimpleXML dependency
### Description

**Is your task request related to a problem? Please describe.**

WordPress core uses SimpleXML, but it's not a required extension. For example, code such as `class_exists( 'SimpleXMLElement' )` can be found in several places in core.

The Core Sitemaps plugin declares `ext-simplexml` as a dependency.

Prior to any merge proposal with core, it needs to be determined whether SimpleXML is a required extension, whether PHP 5.6 as a minimum affects this, behaviour if the extension doesn't exist, etc etc.

swissspidy:
### Issue Number

Fixes #122.

### Description

Uses `wp_die()` when the SimpleXML extension is not available.

Screenshot 2020-03-05 at 15 49 04

### Type of change

### Steps to test

Disable extension or temporarily tweak code to trigger the error message.

Not sure how feasible/worthwhile a CI integration is. But it might be possible to disable the extension there:

https://docs.travis-ci.com/user/languages/php/#disabling-preinstalled-php-extensions

### Acceptance criteria

### DIFF
--- a/inc/class-core-sitemaps-renderer.php
+++ b/inc/class-core-sitemaps-renderer.php
@@ -92,7 +92,12 @@ class Core_Sitemaps_Renderer {
 		header( 'Content-type: application/xml; charset=UTF-8' );
 
 		if ( ! class_exists( 'SimpleXMLElement' ) ) {
-			add_filter( 'wp_die_handler', static function() { return '_xml_wp_die_handler'; } );
+			add_filter(
+				'wp_die_handler',
+				static function () {
+					return '_xml_wp_die_handler';
+				}
+			);
 
 			wp_die(
 				sprintf(
@@ -100,9 +105,9 @@ class Core_Sitemaps_Renderer {
 					__( 'Could not generate XML sitemap due to missing %s extension', 'core-sitemaps' ),
 					'SimpleXML'
 				),
-				__( 'WordPress &rsaquo; Error' ),
+				__( 'WordPress &rsaquo; Error', 'core-sitemaps' ),
 				array(
-					'response' => 501 // Not implemented
+					'response' => 501, // "Not implemented".
 				)
 			);
 		}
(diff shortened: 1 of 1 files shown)


inc/class-core-sitemaps-renderer.php
```diff
@@ -91,6 +91,22 @@ public function get_sitemap_index_stylesheet_url() {
 	public function render_index( $sitemaps ) {
 		header( 'Content-type: application/xml; charset=UTF-8' );
 
+		if ( ! class_exists( 'SimpleXMLElement' ) ) {
+			add_filter( 'wp_die_handler', static function() { return '_xml_wp_die_handler'; } );
```
swissspidy: It didn't seem to work without this. A regular HTML error was displayed otherwise.

pfefferle: Does Google & Co. use and XML Accept header? https://developer.wordpress.org/reference/functions/wp_is_xml_request/

This might be a feature... If you open it in a browser you get the HTML and Search Engines "see" the XML.

swissspidy: I'd rather explicitly send XML all the time.

When integrating this into core, we'd add a new function like `is_sitemap()`. The `wp_die()` function already contains `is_feed()` checks next to `wp_is_xml_request()`, so we could just amend this with a `is_sitemap()` check.

pfefferle (approved):

inc/class-core-sitemaps-renderer.php
pfefferle: this line seems to throw phpcs errors

swissspidy: Ah yeah, need to add some newlines probably.

pbiron:
Shouldn't the availability of the extension be checked for in `Core_Sitemaps_Renderer::render_sitemap()` as well?

However unlikely it is that a crawler would hit an individual sitemap without going through the index, it would be best to check for the extension there as well.

swissspidy:
@pbiron You are correct!

PR #145, opened 2020-03-10
Remove all traces of lastmod

### LINKED ISSUE #43: Timezone support
### Description

https://github.com/GoogleChromeLabs/wp-sitemaps/issues/35 but affects all sitemaps.

Each sitemap needs to have a `<lastmod>` date listed on the sitemap's index. Current thinking on how to approach this is:

> I think the index sitemap will list the buckets and each bucket is a cpt object so has its own post_modified_gmt which is when it was last updated with content, I don’t think it will get much more complex. That bit is missing currently, and then each bit of content registered with a sitemap provided will update its bucket

https://hmn.slack.com/archives/CNVN3ERGV/p1572873401010900

The timezone for each post currently being used on the posts sitemap is: `'lastmod'    => mysql2date( DATE_W3C, $post->post_modified_gmt, false ),`

**Additional context**
Needs some discussion as to whether honouring the `get_option( 'timezone_string' )` and fallback using `get_option( 'gmt_offset' )` is needed on the sitemap index, or is it overkill for the scope of this project

### Acceptance Criteria
- [ ] Display the last modified date on sitemap.xml for each sitemap listed

### LINKED ISSUE #116: Support custom taxonomies of attachment post type
### Description

#22 

Custom taxonomies of attachment post type should be supported (at least through filter).

**Additional context**
Currently, fetching `last_mod` date for custom taxonomies of attachment post type fails and generates PHP notices like:
```
PHP Notice:  Undefined offset: 0 in /[...]/plugins/wp-sitemaps/inc/class-core-sitemaps-taxonomies.php on line 95
```

I believe the problem is that the `get_posts()` method as used [here](https://github.com/GoogleChromeLabs/wp-sitemaps/blob/master/inc/class-core-sitemaps-taxonomies.php#L74) does not account for `attachment` post type as only posts with `post_status = 'public'` are fetched by default, while attachments have `post_status = 'inherit'`. As a result, no posts are fetched, the PHP notice is generated [here](https://github.com/GoogleChromeLabs/wp-sitemaps/blob/master/inc/class-core-sitemaps-taxonomies.php#L95) and the extraction of `last_mod` date fails.

Possible fix would be to check whether current taxonomy is of attachment post type and then explicitly set `post_type` argument in the call to get_posts() to `attachment`. This way [the special treatment](https://core.trac.wordpress.org/browser/tags/5.3/src/wp-includes/post.php#L2019) kicks in and everything works as expected.

### Acceptance Criteria

I'm sorry, but I'm not sure what is expected here.

### LINKED ISSUE #133: Consider removing lastmod field
As discussed on Slack last week, calculating the last modified date for terms is rather difficult and computationally expensive in WordPress. Given that sitemaps are first and foremost a discovery mechanism for content, having this data is not necessarily required.

We should explore omitting this functionality, either completely or only for certain providers.

swissspidy:
Fixes #133.
Fixes #116.
Fixes #43.

To do:

- [x] Documentation in readme.

### Type of change

### Steps to test
Describe the tests required to verify your changes.
Provide instructions so the PR Tester can check functionality and also list any relevant details and / or dependencies required for your tests.

### Acceptance criteria

### DIFF
--- a/inc/class-core-sitemaps-provider.php
+++ b/inc/class-core-sitemaps-provider.php
@@ -91,16 +91,6 @@ class Core_Sitemaps_Provider {
 		 * Shows only on the first page if the reading settings are set to display latest posts.
 		 */
 		if ( 'page' === $type && 1 === $page_num && 'posts' === get_option( 'show_on_front' ) ) {
-			// Assumes the homepage last modified date is the same as the most recent post.
-			$last_modified = get_posts(
-				array(
-					'numberposts'            => 1,
-					'no_found_rows'          => true,
-					'update_post_term_cache' => false,
-					'update_post_meta_cache' => false,
-				)
-			);
-
 			// Extract the data needed for home URL to add to the array.
 			$url_list[] = array(
 				'loc'     => home_url(),
--- a/inc/class-core-sitemaps-provider.php
+++ b/inc/class-core-sitemaps-provider.php
@@ -46,31 +46,6 @@ class Core_Sitemaps_Provider {
 	public $slug = '';
 
 	/**
-	 * Set up relevant rewrite rules, actions, and filters.
-	 */
-	public function setup() {
-		// Set up async tasks related to calculating lastmod data.
-		add_action( 'core_sitemaps_calculate_lastmod', array( $this, 'calculate_sitemap_lastmod' ), 10, 3 );
-		add_action( 'core_sitemaps_update_lastmod_' . $this->slug, array( $this, 'update_lastmod_values' ) );
-
-		if ( ! wp_next_scheduled( 'core_sitemaps_update_lastmod_' . $this->slug ) && ! wp_installing() ) {
-
-			/**
-			 * Filter the recurrence value for updating sitemap lastmod values.
-			 *
-			 * @since 0.1.0
-			 *
-			 * @param string $recurrence How often the event should subsequently recur. Default 'twicedaily'.
-			 *                           See wp_get_schedules() for accepted values.
-			 * @param string $type       The object type being handled by this event, e.g. posts, taxonomies, users.
-			 */
-			$lastmod_recurrence = apply_filters( 'core_sitemaps_lastmod_recurrence', 'twicedaily', $this->slug );
-
-			wp_schedule_event( time(), $lastmod_recurrence, 'core_sitemaps_update_lastmod_' . $this->slug );
-		}
-	}
-
-	/**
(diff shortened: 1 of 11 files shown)


joemcgill:
Related to #133.

swissspidy:
Added a note in the readme now.

inc/class-core-sitemaps-provider.php
```diff
@@ -129,14 +104,12 @@ public function get_url_list( $page_num, $type = '' ) {
...
 			$url_list[] = array(
 				'loc'     => home_url(),
-				'lastmod' => mysql2date( DATE_W3C, $last_modified[0]->post_modified_gmt, false ),
 			);
 		}
 
 		foreach ( $posts as $post ) {
 			$url_list[] = array(
 				'loc'     => get_permalink( $post ),
-				'lastmod' => mysql2date( DATE_W3C, $post->post_modified_gmt, false ),
```
pfefferle, line 139: Shouldn't we keep the lastmod for the objects that support modification dates like Posts/Pages/...?

swissspidy: What's the benefit of the post object type displaying lastmod, but not any other object type? Especially since it wouldn't be displayed on the sitemap index? I'd opt for consistency here.

svandragt: I assumed we concluded that there is no significant search engine benefit, and that was the primary reason to remove it, in which case even if it's easy the feature is smaller by removing lastmod altogether.

Best evidence I could find is: https://webmasters.stackexchange.com/questions/92507/how-important-is-it-to-include-lastmod-in-a-sitemap

>even if the lastmod tag shows the content hasn't been updated since the last crawl the crawler will still crawl it to confirm the content hasn't changed

ThierryA (changes requested):
There are a 7 instances of `$last_modified` (ran a search within the entire plugin) which I believe should be removed.

ThierryA (approved):
LGTM

pbiron:
FYI: the cron jobs to compute `lastmod` (e.g., `core_sitemaps_update_lastmod_users`, etc) still exist in the `cron` option after this PR was merged.

Do we need to add some sort of `update` function in the next release to remove the existing cron entries?

ThierryA:
Given that this is a feature plugin, I don't think it is necessary to add code which will not make it to Core.

pbiron:
WFM...just thought I'd ask.

PR #150, opened 2020-04-07
Move rewrite rule logic out of provider

### LINKED ISSUE #147: maybe flush rewrite rules when a custom provider is added
When registering a custom provider I needed to manually flush rewrite rules before accessing the sitemap generated by that provider.

**To Reproduce**
1. register a custom provider with `core_sitemaps_register_providers`
2. go to `/wp-sitemap.xml`
3. click on the link for the custom provider
4. See the 404

**Expected behavior**
Step 3 should result in the custom provider's sitemap being rendered.

**Additional context**
I discovered this before updating to 0.2.0.  Haven't tested it since updating, but will.

swissspidy:
### Issue Number
Fixes #147

### Description

_todo_

### Type of change

### Steps to test

_todo_

### Acceptance criteria

### DIFF
--- a/inc/class-core-sitemaps.php
+++ b/inc/class-core-sitemaps.php
@@ -112,8 +112,13 @@ class Core_Sitemaps {
 
 		// Register routes for providers.
 		add_rewrite_rule(
-			'^wp-sitemap-([A-z]+)-(([A-z]+)-)?([0-9]+)?\.xml$',
-			'index.php?sitemap=$matches[1]&sitemap-sub-type=$matches[3]&paged=$matches[4]',
+			'^wp-sitemap-([a-z]+?)-([a-z-\d]+?)-([0-9]+?)\.xml$',
+			'index.php?sitemap=$matches[1]&sitemap-sub-type=$matches[2]&paged=$matches[3]',
+			'top'
+		);
+		add_rewrite_rule(
+			'^wp-sitemap-([a-z]+?)-([0-9]+?)\.xml$',
+			'index.php?sitemap=$matches[1]&paged=$matches[2]',
 			'top'
 		);
 	}
--- a/inc/class-core-sitemaps.php
+++ b/inc/class-core-sitemaps.php
@@ -100,7 +100,7 @@ class Core_Sitemaps {
 	public function register_rewrites() {
 		// Add rewrite tags.
 		add_rewrite_tag( '%sitemap%', '([^?]+)' );
-		add_rewrite_tag( '%sub_type%', '([^?]+)' );
+		add_rewrite_tag( '%sitemap-sub-type%', '([^?]+)' );
 
 		// Register index route.
 		add_rewrite_rule( '^wp-sitemap\.xml$', 'index.php?sitemap=index', 'top' );
@@ -112,8 +112,8 @@ class Core_Sitemaps {
 
 		// Register routes for providers.
 		add_rewrite_rule(
-			'^wp-sitemap-([A-z]+)-?([A-z]+)-?([0-9]+)?\.xml$',
-			'index.php?sitemap=$matches[1]&sitemap-sub-type=$matches[2]&paged=$matches[3]',
+			'^wp-sitemap-([A-z]+)-(([A-z]+)-)?([0-9]+)?\.xml$',
+			'index.php?sitemap=$matches[1]&sitemap-sub-type=$matches[3]&paged=$matches[4]',
 			'top'
 		);
 	}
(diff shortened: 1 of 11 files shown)


inc/class-core-sitemaps.php
```diff
@@ -110,11 +111,11 @@ public function register_rewrites() {
...
 
 		// Register routes for providers.
-		$providers = core_sitemaps_get_sitemaps();
-
-		foreach ( $providers as $provider ) {
-			add_rewrite_rule( $provider->route, $provider->rewrite_query(), 'top' );
-		}
+		add_rewrite_rule(
+			'^wp-sitemap-([A-z]+)-?([A-z]+)-?([0-9]+)?\.xml$',
+			'index.php?sitemap=$matches[1]&sitemap-sub-type=$matches[2]&paged=$matches[3]',
```
pfefferle, line 116: What about also rename the query param to `wp-sitemap` with this MR?

swissspidy: Query vars are not typically prefixed in WP. Don't think we should make an exception here.

inc/class-core-sitemaps.php
```diff
@@ -110,11 +111,11 @@ public function register_rewrites() {
 		add_rewrite_rule( '^wp-sitemap-index\.xsl$', 'index.php?sitemap-stylesheet=index', 'top' );
 
 		// Register routes for providers.
-		$providers = core_sitemaps_get_sitemaps();
-
-		foreach ( $providers as $provider ) {
-			add_rewrite_rule( $provider->route, $provider->rewrite_query(), 'top' );
-		}
+		add_rewrite_rule(
+			'^wp-sitemap-([A-z]+)-?([A-z]+)-?([0-9]+)?\.xml$',
```
joemcgill: What permalink would return the users sitemap using this rewrite rule, since there is no sub-type for user objects?

I think maybe this needs to make the second group optional, like:
`^wp-sitemap-([A-z]+)-([A-z]+)?-?([0-9]+)?\.xml$`

joemcgill (commented):
Just leaving notes here for posterity after our weekly chat. I think that we should retain the use of plurals for sitemap types so the paths are like `wp-sitemap-posts-post` rather than `wp-sitemap-post-post`. I _think_ you can probably just update the main `$object_type` values being set in the constructor of each provider.

swissspidy:
@joemcgill I think the latest commit addresses your points

joemcgill:
@swissspidy these changes look better, but I just realized that we also need to be able to support numbers in type/subtype slugs, so we can support taxonomy terms like `best-of-2020`, etc.

swissspidy:
@joemcgill why terms though? we only use taxonomy names like `category` or `post_tag` in the URLs. Trying to come up with a good regex...

inc/class-core-sitemaps.php
```diff
@@ -110,11 +111,16 @@ public function register_rewrites() {
 		add_rewrite_rule( '^wp-sitemap-index\.xsl$', 'index.php?sitemap-stylesheet=index', 'top' );
 
 		// Register routes for providers.
-		$providers = core_sitemaps_get_sitemaps();
-
-		foreach ( $providers as $provider ) {
-			add_rewrite_rule( $provider->route, $provider->rewrite_query(), 'top' );
-		}
+		add_rewrite_rule(
+			'^wp-sitemap-([a-z]+?)-([a-z-\d]+?)-([0-9]+?)\.xml$',
```
pbiron: Minor suggestion for the `sub-type` part of the regex, as stated in the [PHP manual](https://www.php.net/manual/en/regexp.reference.character-classes.php):

> If a minus character is required in a class, it must be escaped with a backslash or appear in a position where it cannot be interpreted as indicating a range, typically as the first or last character in the class. 

So, I think the regex would be more clear/unambiguous if it were:

`^wp-sitemap-([a-z]+?)-([a-z\d-]+?)-([0-9]+?)\.xml$`

or

`^wp-sitemap-([a-z]+?)-([-a-z\d]+?)-([0-9]+?)\.xml$`

I usually use the later, as I think it's easier to see the `-` when it's at the beginning of the character class.

tests/phpunit/sitemaps-renderer.php
```diff
@@ -50,19 +50,19 @@ public function test_get_sitemap_index_stylesheet_url_pretty_permalinks() {
 	public function test_get_sitemap_index_xml() {
 		$entries = array(
 			array(
-				'loc'     => 'http://' . WP_TESTS_DOMAIN . '/wp-sitemap-posts-post-1.xml',
```
pbiron, line 53: These changes in the unit test should also be reverted (including those for the `$expected` value below).  They pass as is, but  think it would be better for long term maintenance of the unit tests if the convention used in this test were what was used for the actual sitemap index.

pbiron (changes requested):

inc/class-core-sitemaps.php
swissspidy: Sounds good!

I'll also replace `[0-9]` at the end with `[\d]` while I'm at it.

tests/phpunit/sitemaps-renderer.php
swissspidy: Agreed!

pbiron (changes requested):
One additional thing related to the unit tests:

the `$route` and `$slug` properties should be removed from `Core_Sitemaps_Test_Provider` in `tests\phpunit\inc\class-core-sitemaps-test-provider.php`.

After that I think this is good to go.

swissspidy:
@pbiron done!

pbiron (approved):
Great work on this @swissspidy!

PR #171, opened 2020-05-01
Make registry API methods consistent in naming

felixarntz:
### Description
Currently the API exposed by `Core_Sitemaps_Registry` is inconsistent:
* `add_sitemap`
* `get_provider`
* `get_sitemaps`

In all cases, it's about registering or getting registered sitemap provider objects, so these names should reflect that. In this PR I'm opting for renaming `get_provider` to `get_sitemap` because I think the "provider" name is more technical and is explained when it comes to method parameters and return values anyway. I'm not opposed to naming everything "provider"-based instead if there's a good reason for that - the main takeaway is that these should be consistently named.

### Acceptance criteria

### DIFF
--- a/inc/class-core-sitemaps-registry.php
+++ b/inc/class-core-sitemaps-registry.php
@@ -48,7 +48,7 @@ class Core_Sitemaps_Registry {
 	}
 
 	/**
-	 * Returns a single registered sitemaps.
+	 * Returns a single registered sitemaps provider.
 	 *
 	 * @since 5.5.0
 	 *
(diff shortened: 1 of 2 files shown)


swissspidy (approved):

inc/class-core-sitemaps-registry.php
```diff
@@ -42,13 +42,13 @@ public function add_sitemap( $name, $provider ) {
 	}
 
 	/**
-	 * Returns a single sitemap provider.
+	 * Returns a single registered sitemaps.
```
joemcgill: This should probably read either:

1. Returns a single registered sitemaps provider.
2. Returns a single registered sitemap.

joemcgill (approved):
One small docs suggestion, but this looks good to me.

inc/class-core-sitemaps-registry.php
felixarntz: Done, I went with 1.

PR #172, opened 2020-05-01
Remove some easily avoidable coupling between classes and WP details

### LINKED ISSUE #103: Add helper function for getting all sitemaps for the index.
### Description

This is a follow-up to #102.

The implementation for gathering all sitemap entries for the sitemap index is implemented procedurally within the `Core_Sitemaps_Index::render_index()` method, and repeated in the `_get_sitemap_entries()` helper method in the PHPUnit tests.

At present, we've not abstracted this into a separate functional method since it's not reused anywhere else in the codebase, nor would we expect it to be needed for implementations extending or modifying the core functionality, and would introduce backwards compatibility concerns. However, as described in https://github.com/GoogleChromeLabs/wp-sitemaps/pull/102#issuecomment-573209132, abstracting this important functionality would make it easier to test and prevent possible future bugs from being introduced, like #102, during code refactoring.

**Additional context**
I would suggest an implementation, similar to the following...

```
function core_sitemaps_get_sitemap_entries() {
	$sitemaps  = array();
	$providers = core_sitemaps_get_sitemaps();

	foreach ( $providers as $provider ) {
		// Using array_push is more efficient than array_merge in a loop.
		array_push( $sitemaps, ...$provider->get_sitemap_entries() );
	}

	return apply_filters( 'filter_name', $sitemaps );
}
```

Note that the use of the [splat operator](https://www.php.net/manual/en/migration56.new-features.php#migration56.new-features.splat) makes this functionality only compatible for PHP 5.6+. To extend compatibility to earlier versions, this could be modified to use `call_user_func_array()` on an array of sitemap entry arrays. See https://github.com/dseguy/clearPHP/blob/master/rules/no-array_merge-in-loop.md for more context.

### Acceptance Criteria
- [ ] A publicly accessible function/method exists that encapsulates the functionality for gathering all sitemap entries for the index, returning all available sitemap pages data.
- [ ] The new method is covered by new PHPUnit tests.
- [ ] The current places in the codebase where this functionality exists has been updated to make use of the abstracted method.

felixarntz:
### Description

This PR bundles a few easy wins to reduce coupling (and partially duplicate functionality) between certain classes and WP-specific details:

1. The `Core_Sitemaps_Index` class represents the sitemap index, but right now it also fires hooks. While these hooks are partially index-related, they are more-so related to WordPress-specific fixes and adjustments which are otherwise handled by the entrypoint `Core_Sitemaps`, where I think those are more feasible. `Core_Sitemaps_Index` is more comparable to `Core_Sitemaps_Provider`, and having it add the side effect of hooks is a bit unexpected I'd argue. See https://github.com/GoogleChromeLabs/wp-sitemaps/commit/3657b7d59d521c2f0c39b1618232b225ccfa5863 and https://github.com/GoogleChromeLabs/wp-sitemaps/commit/8011460d141c05c0c3b35d3cf5ef3eb2dbb7f530.
2. `Core_Sitemaps_Provider` has a `get_url_list` method that is used to render the respective sitemap. It would make sense for `Core_Sitemaps_Index` to have an adequate `get_sitemap_list` method that would be used to render the index. Currently the logic for that is within the `Core_Sitemaps` class, bundled with other functionality. Moving the logic into its own `Core_Sitemaps_Index` method brings it to a potentially more appropriate location and also makes it easily testable (test included). Note that in order to do that `Core_Sitemaps_Index` now receives the `Core_Sitemaps_Registry` instance as dependency. See https://github.com/GoogleChromeLabs/wp-sitemaps/commit/e768fc73a9c8b3d3d1ab5f7bf613dff005e7ef2e.
3. `Core_Sitemaps_Stylesheet::render_stylesheet` currently re-checks `get_query_var( 'sitemap-stylesheet' )` although that's already handled right before in `Core_Sitemaps::render_sitemaps`. This can easily be avoided by passing the type of the stylesheet to render into the method from the query variable and as a result has the stylesheet class independent of WP implementation details. See https://github.com/GoogleChromeLabs/wp-sitemaps/commit/a10691ea2267aec082012692b1738e69968a87a7.

As a result of 1. and 3., all general WP hooks as well as WP implementation details like query variables and request handling are now handled solely by `Core_Sitemaps`, decoupling other classes from them.

Note that as a super-minor unrelated change this PR also removes a completely unused property (was already unused prior to the changes made here): See https://github.com/GoogleChromeLabs/wp-sitemaps/commit/ace99fc90b7a2ab2e8cc43606146799c3d003e15.

Fixes #103 (via 2.)

### Acceptance criteria


inc/class-core-sitemaps-stylesheet.php
```diff
@@ -15,25 +15,23 @@
...
-				echo $this->get_sitemap_stylesheet();
-			}
+	public function render_stylesheet( $type ) {
+		header( 'Content-type: application/xml; charset=UTF-8' );
 
-			if ( 'index' === $stylesheet_query ) {
-				// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- All content escaped below.
-				echo $this->get_sitemap_index_stylesheet();
-			}
+		if ( 'xsl' === $type ) {
```
swissspidy: It's odd that the valid stylesheet types are `xsl` and `index`. We should switch to `index` and `sitemap` or something

inc/class-core-sitemaps-index.php
```diff
@@ -14,35 +14,39 @@
...
-		// Add filters.
-		add_filter( 'robots_txt', array( $this, 'add_robots' ), 0, 2 );
-		add_filter( 'redirect_canonical', array( $this, 'redirect_canonical' ) );
+	public function __construct( $registry ) {
+		$this->registry = $registry;
 	}
 
 	/**
-	 * Prevent trailing slashes.
+	 * Gets a sitemap list for the index.
```
swissspidy, line 40: Let's add `@since 5.5.0`

swissspidy: Also applies to all other new methods, of course.

swissspidy (approved):
A couple of comments but otherwise LGTM 👍

inc/class-core-sitemaps-stylesheet.php
felixarntz: Done.

inc/class-core-sitemaps-index.php
felixarntz: Done.

swissspidy:
@felixarntz mind updating the `Test_Core_Sitemaps_Renderer` tests after eeb21070fa6b63fbf4b6b2de24d360a3a7c0fde0?

felixarntz:
@swissspidy Oops, I thought I had looked for all instances. Should be fixed now.

felixarntz:
~~@swissspidy This seems to have broken `master` - investigating...~~ It was unflushed rewrite rules 😂 🤦

PR #174, opened 2020-05-08
More consistent naming, docs and behavior around object types and subtypes

### LINKED ISSUE #91: Clarify variable/property names
### Description

Sitemap providers

There are several places throughout the codebase where the verbiage around type, subtype, slug, object type, etc. is unclear, leading to function signatures where methods are expecting values like `$name` which are actually referring to the object type or subtype. This is confusing and requires developers to understand the implicit design intent of a property or variable rather than using explicit naming to make the intent clear.

We should update the names of properties and functional variables within all providers to denote object type or object sub-type explicitly and remove references to `$name` and `$slug` throughout those systems, unless those variable/property names have an intentional meaning.

### Acceptance Criteria
- []
- []
- []

felixarntz:
### Issue Number
Fixes https://github.com/GoogleChromeLabs/wp-sitemaps/issues/91

### Description
This PR ensures object subtypes are documented properly and correct variable and method names using `object_subtype` are in place.

* Rename `type` / `sub_type` to `object_subtype` (or in few instances `object_type`) where applicable. In the respective provider sub-classes I opted to reference object subtypes in a less abstract way as `post_type` / `taxonomy`.
* Make `Core_Sitemaps_Provider` abstract, there was no purpose in not having it abstract, and it contained some post-specific code and documentation, which should not be the case.
* Fix inconsistency in what the `get_object_subtypes()` method returns: it sometimes was a list of names, sometimes a keyed map of objects. Now it is always the latter, since that allows for fast `isset` checks on keys as well as easy usage of the post-/taxonomy-specific filters.
* Clarify `core_sitemaps_get_max_urls()` parameter and make it mandatory, since there is no point in having it optional, and having it empty would be unexpected for the filter usage.
* Fix object type names to be `post`, `term`, `user`, as already defined by core. The `posts`, `taxonomies`, `users` names have been kept as provider-specific names, e.g. to be used in URLs as they are now. I wonder whether we should go a bit further and eliminate usage of these throughout the codebase so that the _only_ become necessary for URLs (e.g. we could rename the `Core_Sitemaps_Provider::$name` property to `Core_Sitemaps_Provider::$url_query_value` or something). That would simplify the codebase so that we'd always use the object type as provider identifier and not the "arbitrary" strings introduced for URL purposes.
* Fix some inaccurate documentation and use core conventions (e.g. no variable names in return type docs, "taxonomy type" -> "taxonomy", "Default ''." --> "Default empty.").

### Type of change

### Acceptance criteria

### DIFF
--- a/inc/class-core-sitemaps-provider.php
+++ b/inc/class-core-sitemaps-provider.php
@@ -148,6 +148,20 @@ abstract class Core_Sitemaps_Provider {
 		/* @var WP_Rewrite $wp_rewrite */
 		global $wp_rewrite;
 
+		if ( ! $wp_rewrite->using_permalinks() ) {
+			return add_query_arg(
+				// Accounts for cases where name is not included, ex: sitemaps-users-1.xml.
+				array_filter(
+					array(
+						'sitemap'          => $this->object_type,
+						'sitemap-sub-type' => $name,
+						'paged'            => $page,
+					)
+				),
+				home_url( '/' )
+			);
+		}
+
 		$basename = sprintf(
 			'/wp-sitemap-%1$s.xml',
 			implode(
@@ -163,23 +177,7 @@ abstract class Core_Sitemaps_Provider {
 			)
 		);
 
-		$url = home_url( $basename );
-
-		if ( ! $wp_rewrite->using_permalinks() ) {
-			$url = add_query_arg(
-				// Accounts for cases where name is not included, ex: sitemaps-users-1.xml.
-				array_filter(
-					array(
-						'sitemap'          => $this->object_type,
-						'sitemap-sub-type' => $name,
-						'paged'            => $page,
-					)
-				),
-				home_url( '/' )
-			);
-		}
-
-		return $url;
+		return home_url( $basename );
 	}
 
 	/**
--- a/inc/class-core-sitemaps-provider.php
+++ b/inc/class-core-sitemaps-provider.php
@@ -206,9 +206,9 @@ abstract class Core_Sitemaps_Provider {
 		}
 
 		/**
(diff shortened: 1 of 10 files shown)


inc/class-core-sitemaps-provider.php
```diff
@@ -215,6 +212,8 @@ public function get_object_sub_types() {
 		 *
 		 * @link https://github.com/GoogleChromeLabs/wp-sitemaps/pull/72#discussion_r347496750
 		 */
-		return array( false );
+		return array(
+			'' => (object) array( 'name' => '' ),
```
swissspidy, line 216: The docblock above is no longer applicable.

Also, wouldn't an empty array suffice?

swissspidy (approved):
Overall LGTM, with one nit

inc/class-core-sitemaps-provider.php
felixarntz: @swissspidy The comment actually is still applicable to a degree, it's just slightly off in the detail that the value is now not `false`. The doc-block basically states why this should _not_ be an empty array even though that would look like the reasonable thing to return here. There needs to be a value of some sort in there because otherwise sitemap providers without subtypes result in these sitemaps not being registered.

I actually think this is really strange behavior in the way the current code works, and I agree with you that this _should_ normally return an empty array - it just doesn't work here based on what the current code does. I think we should revisit and potentially fix this in a separate issue.

felixarntz: I've updated the content of the block a bit to clarify.

swissspidy: Makes sense. Could you create such an issue?

swissspidy: @felixarntz created #180.

PR #182, opened 2020-05-12
Remove "core" prefix in functions, hooks, and class file names.

adamsilverstein:
### Issue Number
Part of https://github.com/GoogleChromeLabs/wp-sitemaps/issues/164

### Description
Drop core_sitemaps prefix in functions, hooks, and class file names.

Notes: 
* I left the prefix as `sitemaps` which seems to make sense in almost all cases
* I tried to avoid renaming the plugin itself (and it's root file).
* I kept some slugs that don't change (like the core slack channel.
* The text domain will be removed entirely in the core patch.

I verified the plugin works as expected after these changes.

### Screenshots (before and after if applicable)
If your PR includes visual changes include before and after screenshots showing the change.

### Type of change
Please select the relevant options:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Enhancement (change which improves an existing feature. E.g., performance improvement, docs update, etc.)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Steps to test
Describe the tests required to verify your changes.
Provide instructions so the PR Tester can check functionality and also list any relevant details and / or dependencies required for your tests.

### Acceptance criteria
- [ ] My code follows WordPress coding standards.
- [ ] I have performed a self-review of my own code.
- [ ] If the changes are visual, I have cross browser / device tested.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] My changes generate no new warnings.
- [ ] I have added test instructions that prove my fix is effective or that my feature works.

PR #183, opened 2020-05-18
Add filter to modify WP_Query arguments

### LINKED ISSUE #131: Add filter to modify WP_Query arguments

pfefferle:
### Issue Number
fix #131 

### Description
Add filter to modify WP_Query arguments.

### Type of change

### Acceptance criteria


inc/providers/class-core-sitemaps-posts.php
```diff
@@ -68,7 +68,19 @@ public function get_url_list( $page_num, $post_type = '' ) {
...
+		 * Allows modification of the URL list query arguments before querying.
+		 *
+		 * @see WP_Query for a full list of arguments
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array $args An array of WP_Query arguments.
+		 */
+		$args = apply_filters(
+			'core_sitemaps_post_url_list_query_args',
```
swissspidy: Depending on whether #182 gets in first, we'd need to rename this appropriately.

inc/providers/class-core-sitemaps-posts.php
```diff
@@ -133,7 +147,19 @@ public function max_num_pages( $post_type = '' ) {
...
+		 * Allows modification of the "maximum number of pages" query arguments before querying.
+		 *
+		 * @see WP_Query for a full list of arguments
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array $args An array of WP_Query arguments.
+		 */
+		$args = apply_filters(
+			'core_sitemaps_post_type_max_num_pages_query_args',
```
swissspidy: Maybe `core_sitemaps_posts_max_num_pages_query_args`?

inc/providers/class-core-sitemaps-posts.php
```diff
@@ -133,7 +147,19 @@ public function max_num_pages( $post_type = '' ) {
...
+		/**
+		 * Filters the query arguments for calculating the maximum number of pages.
+		 *
+		 * Allows modification of the "maximum number of pages" query arguments before querying.
+		 *
+		 * @see WP_Query for a full list of arguments
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array $args An array of WP_Query arguments.
```
swissspidy: Let's add `$post_type` as second param

pfefferle: I thought about that too!

inc/providers/class-core-sitemaps-posts.php
pfefferle: I removed it first, but then the PHPCS checks failed, so I re-added the `core_` prefix. Perhaps we should remove the check from the phpcs config.

swissspidy (commented):
Let's also add a filter for the `WP_Term_Query` args in `\Core_Sitemaps_Taxonomies::get_url_list()`.

inc/providers/class-core-sitemaps-posts.php
swissspidy: That's what #182 is dealing with.

inc/providers/class-core-sitemaps-taxonomies.php
```diff
@@ -77,21 +77,35 @@ public function get_url_list( $page_num, $taxonomy = '' ) {
...
+		/**
+		 * Filters the taxonomy terms query arguments.
+		 *
+		 * Allows modification of the taxonomy query arguments before querying.
+		 *
+		 * @see WP_Term_Query for a full list of arguments
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array $args An array of WP_Term_Query arguments.
```
swissspidy: Let's add `$taxonomy` as second param

inc/providers/class-core-sitemaps-posts.php
pfefferle: should I remove it?

swissspidy: No, you can leave it as-is.

What I wanted to say is that we need to rename it if and only if #182 gets merged before this PR.

inc/providers/class-core-sitemaps-posts.php
```diff
@@ -68,7 +68,20 @@ public function get_url_list( $page_num, $post_type = '' ) {
...
+		 *
+		 * @see WP_Query for a full list of arguments
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array  $args      An array of WP_Query arguments.
+		 * @param string $post_type The post type string.
+		 */
+		$args = apply_filters(
+			'core_sitemaps_post_url_list_query_args',
```
felixarntz: In accordance with other filter names (including the other one introduced here below), it'd be more consistent to use the "posts" plural here (`core_sitemaps_posts_url_list_query_args`).

inc/providers/class-core-sitemaps-taxonomies.php
```diff
@@ -77,21 +77,37 @@ public function get_url_list( $page_num, $taxonomy = '' ) {
...
+		 *
+		 * @see WP_Term_Query for a full list of arguments
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array  $args     An array of WP_Term_Query arguments.
+		 * @param string $taxonomy The taxonomy string.
+		 */
+		$args = apply_filters(
+			'core_sitemaps_taxonomy_terms_query_args',
```
felixarntz: Same here, let's use the same plural string "taxonomies" like the other filters do. The other problem here is that the filter name should also be specific to the URL list query, similar like how you did it for posts. In other words, this should be come `core_sitemaps_taxonomies_url_list_query_args`.

inc/providers/class-core-sitemaps-taxonomies.php
```diff
@@ -77,21 +77,37 @@ public function get_url_list( $page_num, $taxonomy = '' ) {
...
+
+				/*
+				* Limits aren't included in queries when hierarchical is set to true (by default).
+				*
+				* @link: https://github.com/WordPress/WordPress/blob/5.3/wp-includes/class-wp-term-query.php#L558-L567
+				*/
+				'hierarchical'           => false,
+				'update_term_meta_cache' => false,
+			),
+			$taxonomy
```
felixarntz: General issue here: If we add a filter for the URL list query args, we should also add a filter for the "max num pages" query args like you did for posts.

I wonder though if we could simplify this, since having to add two filters for each of these things is not a great developer experience. We should be able to instead restructure the code in both the posts and taxonomies provider to use the same array of query arguments, and then just modify the `fields` / `count` / `no_found_rows` arguments _after_ filtering, based on whether we're getting items or counting items.

This is already the case for the users provider, and I think it would be much cleaner if we did this throughout. We can use the same type of `WP_Query` / `WP_Term_Query` instances respectively. How about we introduce a `get_public_terms( $taxonomy )` method here that returns the common "base" query args? We could then have a single `core_sitemaps_taxonomies_query_args` filter there to cover both areas.

inc/providers/class-core-sitemaps-users.php
```diff
@@ -98,14 +98,28 @@ public function get_public_post_authors_query( $page_num = 1 ) {
...
+		 * Allows modification of the authors query arguments before querying.
+		 *
+		 * @see WP_User_Query for a full list of arguments
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array $args An array of WP_User_Query arguments.
+		 */
+		$args = apply_filters(
+			'core_sitemaps_user_query_args',
```
felixarntz: Following up on the previous comment, this filter is much nicer to use than the ones currently in place for posts and taxonomies. You only need to modify this single filter to affect both the users URL list _and_ the users "max num pages" count. Let's apply the same idea for posts and taxonomies.

Only nit-pick here, this should also use plural "users" like elsewhere, i.e. `core_sitemaps_users_query_args`.

inc/providers/class-core-sitemaps-posts.php
```diff
@@ -133,7 +149,20 @@ public function max_num_pages( $post_type = '' ) {
...
+		 *
+		 * @see WP_Query for a full list of arguments
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array  $args      An array of WP_Query arguments.
+		 * @param string $post_type The post type string.
+		 */
+		$args = apply_filters(
+			'core_sitemaps_posts_max_num_pages_query_args',
```
felixarntz: See my other comment related to taxonomies, it would be much better developer experience if we had a single `core_sitemaps_posts_query_args` filter - we could accomplish this by introducing a `get_public_posts( $post_type )` method that returns the common "base" query args.

felixarntz (dismissed):
There are a few slight naming inconsistencies that should be resolved, but actually the current filters being introduced by this PR currently are a bit too complicated to use. We should try to have a single filter for each provider (posts, taxonomies, users) that covers both the "URL list" and the "max num pages". See individual comments for more details.

inc/providers/class-core-sitemaps-posts.php
pfefferle: Now that I think about that... what about removing the filter completely... the "max num" is already part of the filterable `$args` array, so there is no need to filter it at all... and it is also compatible with the other filter...

Rajesh-Techblissonline: I think the query args filter is unnecessary. The query args should fetch only the post ids using "fields" and leave it to plugins to do the filter using the post ids. The plugins can include or exclude post ids based on user selection. A similar approach should be adopted to taxonomy query args. These are better approaches as the plugins should anyway have a list of Post IDs or Term IDs to do the necessary filter. 

https://github.com/GoogleChromeLabs/wp-sitemaps/issues/187

pfefferle: @Rajesh-Techblissonline but some might want to change the query... that’s what the mr is all about!

Rajesh-Techblissonline: @pfefferle That is fine though fetching whole POST instead of just POST IDs might be inefficient as it will require more memory. May be plugin authors can just make the query to retrieve POST IDs alone using the query filter but inefficiencies might creep into some plugins if they retrieve whole POSTs. 

But what is passed to the filter core_sitemaps_posts_url_list towards the end is url_list and if plugin authors like to add additional post attributes like lastmod or change frequency, he will need Post IDs.  A filter core_sitemaps_post_attributes within the POST loop will be better for him to add those attributes. Currently, the function adds loc post attribute to the array and making a filter for the array as shown below will let plugin authors to add their post attributes to the array.

A filter like this as suggested in [https://github.com/GoogleChromeLabs/wp-sitemaps/issues/187 ](url) will be better:

foreach ( $posts as $post ) {
		    $post_sitemap_attributes = array( 'loc' => get_permalink( $post )) ;
        	$url_list[] = apply_filters( 'core_sitemaps_post_attributes', $post_sitemap_attributes, $post );
	    }

This will let plugin authors to add their. $post_sitemap_attributes.

Rajesh-Techblissonline: @pfefferle core_sitemaps_posts_url_list will just give plugin authors a list of URLs and he will have to then retrieve other post_sitemap_attributes using an array of URLs and it will be inefficient.  Allowing them a filter within the POST loop will let him fetch other attributes using the POST ID.

swissspidy: That‘s a completely different use case. Please open a new issue for such an enhancement.

Rajesh-Techblissonline: @swissspidy sure. I will raise a separate issue for that.

When it comes to this issue, if we allow a filter to modify Query arguments, plugin developers should make sure that they either retrieve entire POST, which is inefficient,  or POST ID along with other POST attributes they may require. But retrieving those additional POST attributes may not be necessary as there is no way for them to use those attributes further in the current function. However, POST ID is a must as it is needed to construct the url_list[] array using POST attributes. 

But as it stands now, a plugin author cannot add any additional POST attributes to the url_list array at the individual POST (or POST ID) level within the For Loop. I will raise an issue (enhancement) describing the need for a filter within the loop, to address this use case.

The advantage of allowing a filter for the WP_Query arguments is plugin authors can exclude some POSTS (POST IDs), which is the main use-case for this filter, and also decide the Order of retrieval ( ASC or DESC ).

inc/providers/class-core-sitemaps-taxonomies.php
swissspidy: `get_public_terms_query` perhaps? More in line with `\Core_Sitemaps_Users::get_public_post_authors_query`. Or ideally use the same method name for all providers, i.e. `get_items_query_args`.

Also, should those methods be protected?

inc/providers/class-core-sitemaps-posts.php
pfefferle: @swissspidy should the `$args` array for `get_url_list` and `max_num_pages` be the same?

inc/providers/class-core-sitemaps-taxonomies.php
pfefferle: @swissspidy should the `$args` array for `get_url_list` and `max_num_pages` be the same?

pfefferle: I would vote for the second... it provides more consistence (like `max_num_pages`) and the differentiation comes with the class context.

inc/providers/class-core-sitemaps-posts.php
swissspidy: Let's continue this discussion in the other thread above

inc/providers/class-core-sitemaps-taxonomies.php
swissspidy: > should the `$args` array for `get_url_list` and `max_num_pages` be the same?

I think that's what @felixarntz is proposing, yes.

However, `max_num_pages` would then modify the `$args` _afterwards_ to set `fields=>'ids', 'no_found_rows'=>false`, etc.

Example:

```php

public function get_items_query_args() {
 ...
 return apply_filters( '', $args );
}

public function max_num_pages() {
  $args = $this->get_items_query_args();
  $args['fields'] => 'ids'

  ...
}
```

Pros:

Less duplication

Cons:

Not possible to completely filter all args used in `max_num_pages`

pfefferle: @swissspidy we should also rename `max_num_pages` to `get_max_num_pages` then?

swissspidy: Yeah I think that makes sense.

pfefferle: should I rename them as part of this MR?

pfefferle: it's not that easy to encapsulate the `get_items_query_args` in a generic/abstract function (like `max_num_pages`), because the queries have different params, like for example `offset` for `WP_Term_Query` and `paged` for `WP_Query`.

swissspidy: Oh, I think each provider would implement this on their own, no? It wouldn't live in the abstract provider class.

felixarntz: I don't think we should abstract the methods into like the base provider class, I think it makes sense for these methods to remain in the respective specific provider (I think protected methods make sense because they're implementation details for the respective provider). I don't necessarily think we need to use abstract method names because these methods are specific. How about `get_posts_query_args()` and `get_terms_query_args()`?

felixarntz: @swissspidy 
> Not possible to completely filter all args used in `max_num_pages`

Would this actually be a problem though? I'd argue not being able to filter arguments like `count` is actually cleaner because changing this would likely mess things up anyway by e.g. changing the return type of the query. I'd also say we shouldn't include any pagination arguments in the query, these should also get applied after filtering. What these filters should be used for is modifying the query details for what is queried overall, but not how we paginate or count items.

swissspidy:
@pfefferle @felixarntz Took the liberty to implement some if the feedback added here. PTAL!

inc/providers/class-wp-sitemaps-posts.php
```diff
@@ -141,19 +155,67 @@ public function max_num_pages( $post_type = '' ) {
...
+		 *
+		 * Passing a non-null value will effectively short-circuit the generation,
+		 * returning that value instead.
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param int $max_num_pages The maximum number of pages. Default null.
+		 * @param string $post_type Post type name.
+		 */
+		$max_num_pages = apply_filters( 'wp_sitemaps_taxonomies_pre_max_num_pages', null, $post_type );
```
pfefferle: this should be `wp_sitemaps_posts_pre_max_num_pages`

inc/providers/class-wp-sitemaps-users.php
```diff
@@ -37,7 +37,31 @@ public function __construct() {
...
+		 * Passing a non-null value will effectively short-circuit the generation,
+		 * returning that value instead.
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array  $url_list The URL list. Default null.
+		 * @param int    $page_num Page of results.
+		 */
+		$url_list = apply_filters(
+			'wp_sitemaps_taxonomies_pre_url_list',
```
pfefferle: should be `wp_sitemaps_users_pre_url_list`

inc/providers/class-wp-sitemaps-taxonomies.php
```diff
@@ -62,27 +62,36 @@ public function get_url_list( $page_num, $taxonomy = '' ) {
...
-
-			/*
-			 * Limits aren't included in queries when hierarchical is set to true (by default).
-			 *
-			 * @link: https://github.com/WordPress/WordPress/blob/5.3/wp-includes/class-wp-term-query.php#L558-L567
-			 */
-			'hierarchical'           => false,
-			'update_term_meta_cache' => false,
-		);
+		$args = $this->get_terms_query_args( $taxonomy );
```
pfefferle: I would prefer `get_taxonomies_query_args` because it seems more consistent with the other functions

inc/providers/class-wp-sitemaps-users.php
swissspidy: Whoops! Thanks

inc/providers/class-wp-sitemaps-posts.php
swissspidy: That's what I get for copy/pasting code 😅

pfefferle:
@swissspidy should we also rename `max_num_pages` in `get_max_num_pages` in this MR or should it be a separate one?

swissspidy:
Makes sense here!

swissspidy (approved):
LGTM, but would like an extra sanity check here from @felixarntz or @adamsilverstein

inc/providers/class-wp-sitemaps-posts.php
```diff
@@ -141,19 +155,67 @@ public function max_num_pages( $post_type = '' ) {
...
+		/**
+		 * Filters the query arguments for post type sitemap queries.
+		 *
+		 * @see WP_Query for a full list of arguments.
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array  $args      Array of WP_Query arguments.
+		 * @param string $post_type Post type name.
+		 */
```
Rajesh-Techblissonline, line 203: Shouldn't this check be done?

```php
if ( empty( $post_type ) ) {
			$post_type = $this->get_queried_type();
		}
```

inc/providers/class-wp-sitemaps-posts.php
```diff
@@ -136,24 +150,72 @@ public function get_url_list( $page_num, $post_type = '' ) {
...
+		$args = apply_filters(
+			'wp_sitemaps_posts_query_args',
 			array(
-				'fields'                 => 'ids',
 				'orderby'                => 'ID',
 				'order'                  => 'ASC',
 				'post_type'              => $post_type,
 				'posts_per_page'         => wp_sitemaps_get_max_urls( $this->object_type ),
```
pfefferle, line 209: is there any reason, to not directly call `$this->get_max_num_pages`?

swissspidy: Those are two different things.

`get_max_num_pages` returns the maximum amount of pages.
`wp_sitemaps_get_max_urls` returns the maximum amount of posts on a single page

Not to mention that `get_max_num_pages` calls this function here. You don't wanna cause an infinite loop :-)

pfefferle: Oops, totally misread that! Sorry!

swissspidy: No worries! 👍

inc/providers/class-wp-sitemaps-users.php
```diff
@@ -81,25 +105,41 @@ public function get_url_list( $page_num, $object_subtype = '' ) {
...
+		 * @param int $max_num_pages The maximum number of pages. Default null.
+		 */
+		$max_num_pages = apply_filters( 'wp_sitemaps_users_pre_max_num_pages', null );
+
+		if ( null !== $max_num_pages ) {
+			return $max_num_pages;
+		}
+
+		$args = $this->get_users_query_args();
+		$args['paged'] = 1;
```
felixarntz: Is this actually needed here?

felixarntz (commented):
Mostly LGTM, just one general follow-up question: What's the use-case for all the `pre`-filters? Wouldn't the query filters be sufficient to cover everything?

Plus one specific comment, see below.

inc/providers/class-wp-sitemaps-users.php
swissspidy: I just copied this but... seems like it can be removed, yeah.

swissspidy:
> What's the use-case for all the pre-filters? Wouldn't the query filters be sufficient to cover everything?

I think we discussed this in a meeting or somewhere. It would allow someone to short-circuit sitemap generation and void any database queries. For example because they get the posts from ElasticSearch or somewhere else

felixarntz:
@swissspidy Okay, that's fair. I think we should communicate it this way though, that for most cases the intention is to use the query filters which are simpler to use, while the other 6 ones are for more advanced customizations like not using WordPress queries at all.

Also we'll need new examples for these that should be added to the readme. Maybe in a separate issue/PR?

pfefferle:
@felixarntz we discussed it here https://wordpress.slack.com/archives/CTKTGNJJW/p1589898060283600

PR #184, opened 2020-05-19
Add `wp_sitemaps_index_entry` filter

swissspidy:
### Issue Number

See #151.
See #201.

### Description

Adds a new `wp_sitemaps_index_entry` filter to complement the existing `wp_sitemaps_$provider_entry` filters.

Could be used to add `lastmod` to the index.

### Type of change

### Steps to test

Use the `wp_sitemaps_index_entry` filter to verify expected output.

### Acceptance criteria


composer.json
```diff
@@ -61,6 +61,9 @@
     "ext-simplexml": "*"
   },
   "require-dev": {
+    "ext-dom": "*",
```
swissspidy, line 64: These were just added to make my IDE happy.

inc/class-core-sitemaps-index.php
```diff
@@ -53,7 +53,14 @@ public function get_sitemap_list() {
...
 
-		return $sitemaps;
+		/**
+		 * Filters the array of URLs for a sitemap index before rendering.
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array $sitemaps Array of URLs for a sitemap index.
+		 */
+		return apply_filters( 'core_sitemaps_sitemap_list', $sitemaps );
```
swissspidy: In case someone wanted to add back `lastmod` to the index

inc/class-core-sitemaps-renderer.php
```diff
@@ -150,7 +150,10 @@ public function get_sitemap_index_xml( $sitemaps ) {
 
 		foreach ( $sitemaps as $entry ) {
 			$sitemap = $sitemap_index->addChild( 'sitemap' );
-			$sitemap->addChild( 'loc', esc_url( $entry['loc'] ) );
+			// Add each attribute as a child node to the <sitemap> entry.
+			foreach ( $entry as $attr => $value ) {
```
swissspidy: Loop, because there's no a filter for this list (e.g. to add back `lastmod` if wanted)

inc/class-core-sitemaps-index.php
adamsilverstein: It would be great to document how developers could do this or add `changefreq` or `priority`.

swissspidy: Docs added to the readme!

inc/class-wp-sitemaps-renderer.php
```diff
@@ -195,15 +198,71 @@ public function get_sitemap_xml( $url_list ) {
...
+		/**
+		 * Filters the `<urlset>` attributes for the sitemap.
+		 *
+		 * Can be used to add support for additional namespaces.
+		 *
+		 * @since 5.5.0
+		 *
+		 * @param array $attributes Associative array of urlset attributes and their values.
+		 */
+		$attributes = apply_filters( 'wp_sitemaps_urlset_attributes', $attributes );
```
pbiron: The name of this filter is not ideal.  It is not filtering `attributes`, it is filtering namespace bindings.

ETA: It would be better if it were named `wp_sitemaps_namespace_bindings` or `wp_sitemaps_namespaces`.

inc/class-wp-sitemaps-renderer.php
```diff
@@ -195,15 +198,71 @@ public function get_sitemap_xml( $url_list ) {
...
+
+		foreach ( $attributes as $attribute => $value ) {
+			$urlset->addAttribute( 'xmlns:' . $attribute, $value );
```
pbiron: This is taking advantage of an undocumented "quirk" of SimpleXML.

Notice that callback hooked to `wp_sitemaps_urlset_attributes` in the unit tests (https://github.com/GoogleChromeLabs/wp-sitemaps/pull/184/files#diff-29c6e6f7a1d8eb49cc8790688140547eR398) is uses `xmlns:image` as a key, and hence this is actually doing
```
$urlset->addAttribute( 'xmlns:xmlns:image', ''http://www.google.com/schemas/sitemap-image/1.1' )
```
which is syntactically incorrect for XML, and why I say this usage is a "quirk" in SimpleXML.

The other "quirky" thing about this usage in SimpleXML is that namespaces declared with this "quirk" are **not** returned by `SimpleXMLElement::getDocNamespaces()` nor `SimpleXMLElement::getNamespaces()`.

It would be much better if the keys returned by the filter were "simple" prefixes, e.g.,

```
public function _filter_urlset_attributes( $attributes ) {
	$attributes['image'] = 'http://www.google.com/schemas/sitemap-image/1.1';
	return $attributes;
}
```

And then in `WP_Sitemaps_Renderer::get_sitemap_xml()`, do something like:

```
$nsDecls = array();

/**
 * Filters the `<urlset>` attributes for the sitemap.
 *
 * @since 5.5.0
 *
 * @param array $attributes Associative array of urlset attributes and their values.
 */
$namespace_bindings     = apply_filters( 'wp_sitemaps_urlset_attributes', $nsDecls );
$namespace_bindings[''] = 'http://www.sitemaps.org/schemas/sitemap/0.9';

foreach ( $namespace_bindings as $prefix => $namespace_uri ) {
	if ( $prefix ) {
		$nsDecls[] = "xmlns:{$prefix}=\"{$namespace_uri}\"";
	} else {
		$nsDecls[] = "xmlns=\"{$namespace_uri}\"";
	}
}
$nsDecls = implode( ' ', $nsDecls );

$urlset = new SimpleXMLElement(
	sprintf(
		'%1$s%2$s%3$s',
		'<?xml version="1.0" encoding="UTF-8" ?>',
		$this->stylesheet,
		"<urlset {$nsDecls} />"
	)
);
```

inc/class-wp-sitemaps-renderer.php
```diff
@@ -195,15 +198,71 @@ public function get_sitemap_xml( $url_list ) {
...
+				if ( is_array( $value ) ) {
+					foreach ( $value as $child_attr => $child_value ) {
+						$item = $url->addChild( $prefix . $attr );
+						foreach ( (array) $child_value as $grandchild_attr => $grandchild_value ) {
+							$item->addChild( $prefix . $grandchild_attr, $grandchild_value );
+						}
+					}
 				} else {
-					$url->addChild( $attr, esc_attr( $value ) );
+					$url->addChild( $prefix . $attr, esc_attr( $value ) );
```
pbiron: This is also taking advantage of a similar undocumented "quirk" in SimpleXML.

What this evaluates to at run-time is something like:

```
$url->addChild( 'image:image:loc', esc_attr( $value ) );
```
which is syntactically incorrect for XML, but that's the "quirk" in SimpleXML to allow prefixed elements to added (without specifically passing the namespace URI in the 3rd parameter).

I think it would be **far** better to add a helper method

```
protected function expand_qname( $qname ) {
	if ( false === strpos( $qname, ':' ) ) {
		$local_name = $qname;
		$prefix = '';
	} else {
		list( $prefix, $local_name ) = explode( ':', $qname );
		if ( ! $local_name ) {
			$local_name = $prefix;
			$prefix = '';
		}
	}

	return array( $prefix, $local_name );
}
```

and then do 

```
list( $prefix, $local_name ) = $this->expand_qname( $name );
// $namespace_bindings is array returned by the 'wp_sitemaps_urlset_attributes' filter
// plus $namespace_bindings[''] = 'http://www.sitemaps.org/schemas/sitemap/0.9' added.
$url->addChild( $local_name, esc_attr( $value ), $namespace_bindings[ $prefix ] );
```

inc/class-wp-sitemaps-renderer.php
```diff
@@ -195,15 +198,71 @@ public function get_sitemap_xml( $url_list ) {
...
+				 * </image:image>
+				 * <image:image>
+				 *   <image:loc>http://example.com/image2.jpg</image:loc>
+				 *   <image:title>Cats and Dogs</image:loc>
+				 * </image:image>
+				 */
+				if ( is_array( $value ) ) {
+					foreach ( $value as $child_attr => $child_value ) {
+						$item = $url->addChild( $prefix . $attr );
+						foreach ( (array) $child_value as $grandchild_attr => $grandchild_value ) {
```
pbiron: This is incorrect, since when `$child_value` is a string, `$grandchild_attr` will be `0` and an element with name `0` will be added :-(

Like the helper method mentioned below, it would be easier to add another helper method such as:

```
protected function add_child( $parent_element, $name, $value, $namespace_bindings ) {
	list( $prefix, $local_name ) = $this->expand_qname( $name );

	if ( is_array( $value ) ) {
		$child = $parent_element->addChild( $local_name, '', $namespace_bindings[ $prefix ] );
		foreach ( $value as $child_name => $child_value ) {
			list( $child_prefix, $child_local_name ) = $this->expand_qname( $child_name );

			$this->add_child( $child, $child_local_name, $child_value, $namespace_bindings[ $child_prefix ] );
		}
	} else {
		$parent_element->addChild( $local_name, esc_attr( $value ), $namespace_bindings[ $prefix ] );
	}
}
```

which would allow `grand_grand_children`, etc to be added.

And then the loop would simply become:

```
foreach ( $url_item as $attr => $value ) {
	$this->add_child( $url, $attr, $value );
}
```

pbiron (dismissed):

inc/class-wp-sitemaps-renderer.php
pbiron: just a minor point, but using `$attr` there has always bugged me :-)

Coming at this from an XML perspective, I expect `$attr` to be used for "XML attributes" and this is an element name; hence, it would be easier to read if it were

```
foreach ( $url_item as $name=> $value ) {
	$this->add_child( $url, $name, $value );
}
```

pbiron:
So, I've been thinking about the registry idea that @joemcgill proposed in https://github.com/GoogleChromeLabs/wp-sitemaps/issues/151#issuecomment-620658374 and I really like it over the filter proposed in this PR (and in #151)!!

The main reason is that it will solve the prefix "collision" problem mentioned in https://github.com/GoogleChromeLabs/wp-sitemaps/issues/151#issuecomment-612252959.

That is, suppose plugin X adds Google Image Sitemap properties and so does:

```
add_filter( 'wp_sitemaps_urlset_attributes', function( $attributes ) {
	$attributes['image'] = 'http://www.google.com/schemas/sitemap-image/1.1';
} , 10 );
```

and plugin Y adds completely custom properties but just happens to use the `image` prefix (and just to complicate things, also uses `image` for the local name of the element it adds):

```
// priority 11 is used in this example to guarantee that this binding will "win".
add_filter( 'wp_sitemaps_urlset_attributes', function( $attributes ) {
	$attributes['image'] = 'urn:mydomain.com:image';
} , 11 );
```

That will result in the markup added to the sitemap by plugin X to not be interpreted correctly, because the `image` prefix will be bound to `urn:mydomain.com:image_sitemaps` and not `http://www.google.com/schemas/sitemap-image/1.1` as plugin X intended.

Namespace binding registry functionality can easily be added to the existing `WP_Sitemaps_Registry` class, as in:

```
class WP_Sitemaps_Registry {
	/**
	 * Registered namespace URI -> prefix bindings.
	 *
	 * Keys are namespace URIs, values are prefixes to use for that namespace URI.
	 *
	 * @since 5.5.0
	 *
	 * @var array
	 */
	private $namespace_bindings = array(
		// bind the default namespace to the sitemaps URI.
		'http://www.sitemaps.org/schemas/sitemap/0.9' => '',
	);

...

	/**
	 * Register an XML Namespace URI for use in a sitemap.
	 *
	 * Multiple calls to `register_namespace()` with the same
	 * `$uri` parameter will result in the same prefix being
	 * returned each time.
	 *
	 * @since 5.5.0
	 *
	 * @param string $uri The namespace URI to register.
	 * @returns string    The prefix to use for `$uri`.
	 */
	function register_namespace( $uri ) {
		if ( isset( $this->namespace_bindings[ $uri ] ) ) {
			return $this->namespace_bindings[ $uri ];
		}

		// The generated prefix is arbitrary, as long as it is unique.
		// 'ns' + a number is a convention used in many XML tools.
		$this->namespace_bindings[ $uri ] = 'ns' . count( $this->namespace_bindings );

		return $this->namespace_bindings[ $uri ];
	}

	function get_namespace_prefix( $uri ) {
		if ( isset( $this->namespace_bindings[ $uri ] ) ) {
			return $this->namespace_bindings[ $uri ];
		}

		return false;
	}

	function get_namespace_uri( $prefix ) {
		$namespace_bindings = $this->get_namespace_bindings();
		if ( isset( $namespace_bindings[ $prefix ] ) ) {
			return $namespace_bindings[ $prefix ];
		}

		return false;
	}

	function get_namespace_bindings() {
		return array_flip( $this->namespace_bindings );
	}
}
```

[note that the DocBlocks for all but the `register_namespace()` method have been left out for consiseness]

and then helper functions can be added, such as

```
function wp_sitemaps_register_namespace( $uri ) {
	$sitemaps = wp_sitemaps_get_server();

	return $sitemaps->registry->register_namespace( $uri );
}

...
```

And then plugin X would do:

```
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry, $post, $post_type ) {
	$prefix = wp_sitemaps_register_namespace( 'http://www.google.com/schemas/sitemap-image/1.1' );
	
	$sitemap_entry[ "{$prefix}:image" ] = array();
	foreach ( get_attached_media( 'image', $post ) as $image ) {
		$sitemap_entry[ "{$prefix}:image" ][] = array(
			"{$prefix}:loc"   => get_permalink( $image );
			"{$prefix}:title" => get_the_title( $image );
		);
	}
} );
```

And plugin Y would do:

```
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry, $post, $post_type ) {
	$prefix = wp_sitemaps_register_namespace( 'urn:mydomain.com:image' );
	
	$sitemap_entry[ "{$prefix}:image" ] = array(
		"{$prefix}:id"    => $post->ID,
		"{$prefix}:title" => get_the_title( $post );
	);
} );
```

And that would result in a sitemap such as

```
<urlset
		xmlns='http://www.sitemaps.org/schemas/sitemap/0.9'
		ns1='http://www.google.com/schemas/sitemap-image/1.1'
		ns2='urn:mydomain.com:image'>
	<loc>https://example.com/post-1/</loc>
	<!-- n1:image will be correctly interpreted as a Google Image Sitemaps element -->
	<ns1:image>
		<ns1:loc>https://example.com/wp-content/uploads/2020/05/28/image1.jpg</ns1:loc>
		<ns1:title>Image title</ns1:title>
	</ns1:image>
	<ns1:image>
		<ns1:loc>https://example.com/wp-content/uploads/2020/05/28/image2.jpg</ns1:loc>
		<ns1:title>Another image title</ns1:title>
	</ns1:image>
	<!-- ns2:image will be correctly interpreted as plugin Y's custom image element -->
	<ns2:image>
		<ns2:id>123</ns2:id>
		<ns2:title>My Title</ns2:title>
	</ns2:image>
</urlset>
```

inc/class-wp-sitemaps-renderer.php
swissspidy: When would there ever be a need for grand grand children?

swissspidy:
That looks great, but isn‘t that a bit over-engineered? I don‘t think that‘s a common real world use case for these sitemaps.

inc/class-wp-sitemaps-renderer.php
pbiron: the "need" is determined by whoever defines the markup for the extension.

Since the Sitemaps spec does **not** place an arbitrary max depth on elements in extension namespaces, I don't think core should either, especially since using a recursive helper method (as above) is trivially easy.  Thus, it seems to me that if an arbitrary max depth is to be enforced it should be up to those who want that enforced to argue for why _that_ is necessary.

Am I aware of any published specs for sitemaps extensions that currently require it?  No.

I know Video Sitemaps are out-of-scope **for core**, but consider the following.

Google's description of [Video Sitemaps](https://support.google.com/webmasters/answer/80471?hl=en&ref_topic=4581190) mentions that Google also supports [Media RSS](https://www.rssboard.org/media-rss).  One of the elements defined in Media RSS is [media:location](https://www.rssboard.org/media-rss#media-peerlink).  The documentation for that element says

> Optional element to specify geographical information about various locations captured in the content of a media object.  The format conforms to [geoRSS](http://www.georss.org/).

and then gives the example:

```
<media:location description="My house" start="00:01" end="01:00">
  <georss:where>
    <gml:Point>
      <gml:pos>35.669998 139.770004</gml:pos>
    </gml:Point>
  </georss:where>
</media:location>
```

Granted, there is no equivalent to `<media:location>` in the [published XML Schema](https://www.google.com/schemas/sitemap-video/1.1/sitemap-video.xsd) for Video Sitemaps (and that XML Schema does not allow any extensions), but if Google does **fully** support Media RSS for indexing videos, then it's possible that at some point in the future they will release a version of Video Sitemaps that also supports it...and _that_ would require grand grand children.

Again, I know Video Sitemaps are out-of-scope **for core**, but the whole point of #151 is to allow plugins to support them (in their current and **future** form) without needing to completely duplicate the infrastructure provided by core sitemaps...as long as providing that ability to plugins is not hard and I don't think it is.

pbiron:
> That looks great, but isn‘t that a bit over-engineered?

What's over engineered about a simple array lookup?

pbiron:
> I don‘t think that‘s a common real world use case for these sitemaps.

The example is not a use case, it's simply demonstrating the consequences of the "decision" for plugins to use namespace prefixes when adding extension elements to the sitemap (which, admittedly, was originally my idea in #151).

Since namespace prefixes are **completely arbitrary**, plugin authors should be free to use whatever prefixes they want, including plugin X that adds image sitemap info using `ns1` and plugin Y that adds video sitemap info also using `ns1`.

I think it would be better to **not** use prefixes and to have plugins explicitly specify the namespace URI for their extension elements, but that causes "developer experience" problems of it's own.

For example, instead of using the keys of the sitemap entries to specify the element name, we could use an

```
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry, $post, $post_type ) {
	$sitemap_entry[] = array(
		'uri' => 'http://www.google.com/schemas/sitemap-image/1.1',
		'local_name' => 'image',
		'value' => array(
			array(
				'uri' => 'http://www.google.com/schemas/sitemap-image/1.1',
				'local_name' => 'loc',
				'value' => 'https://example.com/wp-content/uploads/2020/05/29/image1.jpg',
			),
		),
	);
} );
```

but that just seems overly verbose.

My preferred solution would be to use the [URIQualifiedName](https://www.w3.org/TR/xpath-30/#doc-xpath30-URIQualifiedName) syntax from the XPath 3.0 spec that I've used in passing in descriptions/comments of various issues and PRs, e.g.

```
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry, $post, $post_type ) {
	$sitemap_entry['Q{http://www.google.com/schemas/sitemap-image/1.1}image'] = array(
		'Q{http://www.google.com/schemas/sitemap-image/1.1}loc' => 'https://example.com/wp-content/uploads/2020/05/29/image1.jpg';
	);
} );
```

But I suspect plugin authors will just find that "weird", and will too often get the syntax wrong :-(

p.s., extracting the namespace URI and local_name from a `URIQualifiedName` is easy:

```
function expand_URIQualifiedName( $name ) {
	// these regex fragments are from their correpsondingly named productions in
	// the XML, XML Namespaces, and XPath 3.0 specs.
	$NameStartChar    = 'A-Za-z_\x{C0}-\x{D6}\x{D8}-\x{F6}\x{F8}-\x{2FF}\x{370}-\x{37D}\x{37F}-\x{1FFF}\x{200C}-\x{200D}\x{2070}-\x{218F}\x{2C00}-\x{2FEF}\x{3001}-\x{D7FF}\x{F900}-\x{FDCF}\x{FDF0}-\x{FFFD}\x{10000}-\x{EFFFF}';
	$NameChar         = "{$NameStartChar}\-\.0-9\x{B7}\x{0300}-\x{036F}\x{203F}-\x{2040}";
	$NCName           = "([{$NameStartChar}][{$NameChar}]*)";
	$BracedURILiteral = "Q{([^{}]*?)}{$NCName}";
	$URIQualifiedName = "{$BracedURILiteral}|{$NCName}";

	$name    = trim( $name );
	$matches = array();
	if ( preg_match( "@^{$URIQualifiedName}\$@u", $name, $matches ) ) {
		if ( ! isset( $matches[3] ) ) {
			$local_name = $matches[1];
			$uri        = $matches[2];
		} else {
 			$local_name = $matches[3];
 		}
	} else {
		// $local_name being empty signals to the caller that $name is not a URIQualifiedName.
		$local_name = '';
		$uri        = '';
	}

	return array( $local_name, $uri );
}
```

pbiron:
And to further complicate things, none of what is in wp-sitemaps 0.3.0, the proposal in #151, nor this PR provide a way for plugins to add XML attributes in their extension markup :-(

And Video Sitemaps has attributes in it's markup, e.g., [video:restriction](https://support.google.com/webmasters/answer/80471?hl=en&ref_topic=4581190#country-restriction) has a **required** `@relationship` attribute

```
<video:restriction relationship="allow">CA MX</video:restriction>
```

To quote Albert Einstein

> It can scarcely be denied that the supreme goal of all theory is to make the irreducible basic elements as simple and as few as possible without having to surrender the adequate representation of a single datum of experience.

which is often paraphrased as

> Everything should be made as simple as possible, but no simpler

So, **any** extension mechanism we provide really _should_ find a way for plugins to add **arbitrary** XML markup to a sitemap (including XML attributes!) as long as it is in an extension namespace.

swissspidy:
That‘s an excellent point!

Suggestion: leave this for 5.5.x to fully flesh out the details and nail this.

pbiron:
> Suggestion: leave this for 5.5.x to fully flesh out the details and nail this.

I think that's a good suggestion.

So, there are 2 options for what means for 5.5.0:

1. **nothing** other than `<loc>` will allowed
2. we allow plugins to add `<lastmod>`, `<changefreq>` and `<priority>` but nothing else

I'd prefer the 2nd but would be fine with the 1st.

swissspidy:
Let's strive for the 2nd

pbiron:
> Let's strive for the 2nd

#201 is my attempt to do that.

tests/phpunit/sitemaps-renderer.php
```diff
@@ -142,15 +186,15 @@ public function test_get_sitemap_xml() {
 					'</urlset>';
 
 		$this->assertXMLEquals( $expected, $actual, 'Sitemap page markup incorrect.' );
-	}
+	}
```
swissspidy:
```suggestion
	}
```

inc/class-wp-sitemaps-renderer.php
```diff
@@ -198,12 +205,12 @@ public function get_sitemap_xml( $url_list ) {
 		foreach ( $url_list as $url_item ) {
 			$url = $urlset->addChild( 'url' );
 
-			// Add each attribute as a child node to the URL entry.
```
pbiron: This change isn't necessary because it's handled in #201 (https://github.com/GoogleChromeLabs/wp-sitemaps/pull/201/files#diff-cc90739d8286d058f89eaf22de6cec61L201)

inc/class-wp-sitemaps-renderer.php
```diff
@@ -150,7 +150,14 @@ public function get_sitemap_index_xml( $sitemaps ) {
 
 		foreach ( $sitemaps as $entry ) {
```
pbiron, line 151: This should check whether `$name` is `loc` or `lastmod` and doing a `_doing_it_wrong()` if not, as in

```
foreach ( $sitemaps as $entry ) {
	$sitemap = $sitemap_index->addChild( 'sitemap' );
	// Add each element as a child node to the URL entry.
	foreach ( $entry as $name => $value ) {
		if ( 'loc' === $name ) {
			$sitemap->addChild( $name, esc_url( $value ) );
		} elseif ( 'lastmod' === $name ) {
			$sitemap->addChild( $name, esc_attr( $value ) );
		} else {
			_doing_it_wrong(
				__METHOD__,
				/* translators: %s: list of element names */
				sprintf(
					__( 'Fields other than %s are not currently supported for sitemap indexes.', 'core-sitemaps' ),
					implode( ',', array( 'sitemap', 'lastmod' ) )
				),
				'5.5.0'
			);
		}
	}
}
```
```

**Review by pbiron — CHANGES_REQUESTED** [review]
If you want, I can update the index stylesheet to display a column for `lastmod` like #201 does for the provider sitemaps.

File: inc/class-wp-sitemaps-renderer.php (hunk 9 above)
Comment (swissspidy) [reply inc/class-wp-sitemaps-renderer.php]: Not a big deal, will be resolved with the merge

**Comment by swissspidy** [comment]
@pbiron Mind taking another look? Updated the index stylesheet too.

File: inc/class-wp-sitemaps-renderer.php [hunk 11]
```diff
@@ -150,7 +150,25 @@ public function get_sitemap_index_xml( $sitemaps ) {
...
+					$sitemap->addChild( $name, esc_url( $value ) );
+				} elseif ( 'lastmod' === $name ) {
+					$sitemap->addChild( $name, esc_attr( $value ) );
+				} else {
+					_doing_it_wrong(
+						__METHOD__,
+						/* translators: %s: list of element names */
+						sprintf(
+							__( 'Fields other than %s are not currently supported for sitemaps.', 'core-sitemaps' ),
+							implode( ',', array( 'loc', 'lastmod', 'changefreq', 'priority' ) )
```
swissspidy:
```suggestion
							implode( ',', array( 'loc', 'lastmod' ) )
```

inc/class-wp-sitemaps-renderer.php
```diff
@@ -150,7 +150,25 @@ public function get_sitemap_index_xml( $sitemaps ) {
...
+				if ( 'loc' === $name ) {
+					$sitemap->addChild( $name, esc_url( $value ) );
+				} elseif ( 'lastmod' === $name ) {
+					$sitemap->addChild( $name, esc_attr( $value ) );
+				} else {
+					_doing_it_wrong(
+						__METHOD__,
+						/* translators: %s: list of element names */
+						sprintf(
+							__( 'Fields other than %s are not currently supported for sitemaps.', 'core-sitemaps' ),
```
swissspidy:
```suggestion
							__( 'Fields other than %s are not currently supported for the sitemap index.', 'core-sitemaps' ),
```

swissspidy:
@pbiron thanks for the tests! now if you can reply to the bot I believe this is gonna be in good shape

pbiron:
@googlebot I consent

pbiron (approved):

PR #163, opened 2020-04-26
Allow displaying custom columns in stylesheet

pbiron:
### Issue Number

#152

### Description

This is a simplified version of the solution in #153.

Introduces the `core_sitemaps_stylesheet_columns` filter that allows plugins to change the columns rendered by the stylesheet.  This new filter is analogous to `manage_posts_columns` (and all the other `WP_List_Table` "columns" filters).

Unlike #153, the stylesheet generated by this PR does **not** try to render columns for **any** child of `Q{http://www.sitemaps.org/schemas/sitemap/0.9}url`, just those returned by `core_sitemaps_stylesheet_columns`, which means the stylesheet is greatly simplified from the one in that PR...and only _slightly_ more complex than the current stylesheet.

Like #153, it also introduces the `esc_xml()` and `esc_xml__()` functions.

There is a pretty extensive unit test for the rendered HTML.  The `Test_Core_Sitemaps_Renderer::loadXML()` method (introduced in #157) is also modified so that it can load XML from a file, which is needed by `test_stylesheet_html()` (the details of why that is necessary are explained in an inline comment in the test).

The index stylesheet is also cleaned up a bit, although custom columns are **not** allowed for it.  Why?  Because there are currently no hooks that allow plugins to add extension elements to the sitemap index.  If/when those hooks are added, the index stylesheet can easily be adapted to allow custom columns.

#### No custom columns

no-custom-columns

#### Change the text used for the URL column

```
add_filter(
	'core_sitemaps_stylesheet_columns',
	function( $columns ) {
		$columns['http://www.sitemaps.org/schemas/sitemap/0.9']['loc'] = 'Permalink';

		return $columns;
	}
);
```
rename-url

#### Add Change Frequency

```
add_filter(
	'core_sitemaps_posts_url_list',
	function ( $url_list ) {
		foreach ( $url_list as &$url ) {
			$url['changefreq'] = 'weekly';
		}
		
		return $url_list;
	}
);
add_filter(
	'core_sitemaps_stylesheet_columns',
	function( $columns ) {
		$columns['http://www.sitemaps.org/schemas/sitemap/0.9']['changefreq'] = 'Change Frequency';

		return $columns;
	}
);
```
changefreq

#### Reorder columns

```
// I don't know why a plugin would want to do this;
// I also don't know why they'd want to reorder columns
// on WP_Posts_List_Table, but manage_posts_columns allows them to do that :-)
add_filter(
	'core_sitemaps_posts_url_list',
	function ( $url_list ) {
[...]

PR #192, opened 2020-05-26
Adds esc_xml() and esc_xml__() functions.

pbiron:
### Description
These functions were originally introduced in #163.  This simply splits them out into their own PR.

This also introduces a new filter: `esc_xml`.

Both new functions and the new filter are the direct equivlants of `esc_html()`, `esc_html__()` and `esc_html` in core.

The only difference here is that HTML character entities are converted to their Unicode codepoints.

These new functions do **not** 100% guarantee that the resulting text is "safe" for use in an XML instance, because there are certain Unicode codepoints that are forbidding in XML see https://www.w3.org/TR/REC-xml/#charsets (note: the comment about the `Char` production in the XML spec is not completely accurate: many control characters are also not allowed) and it is an open question what to do when the text passed to these functions/filter contain characters which are not allowed in XML.

### Type of change

- [X] New feature (non-breaking change which adds functionality)

### Steps to test
Still needs unit tests

### Acceptance criteria


pbiron:
Once this is merged, then all current uses of `esc_attr()` when adding text context to a sitemap (or index) should be replaced with `esc_xml()` and uses of `__()` for adding text to sitemaps/indexs/stylesheets should be replaced with `esc_xml__()`.

inc/functions.php
```diff
@@ -117,3 +117,51 @@ function core_sitemaps_get_max_urls( $object_type ) {
...
+ */
+function esc_xml( $text ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
+	$safe_text = wp_check_invalid_utf8( $text );
+	$safe_text = _wp_specialchars( $safe_text, ENT_QUOTES );
+	$safe_text = html_entity_decode( $safe_text, ENT_HTML5 );
+	/**
+	 * Filters a string cleaned and escaped for output in XML.
+	 *
+	 * Text passed to esc_xml() is stripped of invalid or special characters
+	 * before output.  HTML named character references are converted to their
```
swissspidy:
```suggestion
	 * before output. HTML named character references are converted to their
```

inc/functions.php
```diff
@@ -117,3 +117,51 @@ function core_sitemaps_get_max_urls( $object_type ) {
...
+	 * equiablent code points.
```
swissspidy:
```suggestion
	 * equivalent code points.
```

pbiron: oops...

pbiron:
In core there is also `esc_html_e()` and `esc_html_x()`.  We don't need XML equivalents of those for sitemaps, but for the core merge they would be good.  @swissspidy  Should I add them to this PR?

swissspidy:
@pbiron For completeness sake, why not 👍

btw, please let me know if the tests I added make sense.

pbiron:
> @pbiron For completeness sake, why not 👍

will do.

> btw, please let me know if the tests I added make sense.

at first glance they look good.  will check in more detail shortly.

swissspidy:
I presume we can also use these new functions already in the existing code base?

pbiron:
> I presume we can also use these new functions already in the existing code base?

Yes, that's what I meant in https://github.com/GoogleChromeLabs/wp-sitemaps/pull/192#issuecomment-634086286

pbiron:
thanx for the unit tests.  They pointed out that calling `html_entity_decode( $safe_text, ENT_HTML5 );` on the entire text is "too aggressive" (i.e., decodes too much, e.g., replaces `&amp;` with `&`).  I think I've got a fix for that, just trying to make it a little more efficient.

Also, `esc_xml()` should not do **any** modifications within [CDATA Sections](https://www.w3.org/TR/REC-xml/#dt-cdsection).  I'll work on that as soon as I make the replacement of HTML named character entities less aggressive.

swissspidy:
> Also, `esc_xml()` should not do any modifications within CDATA Sections

How would you detect CDATA Sections? I'd say if someone uses them, they shouldn't be calling the function? We could suggest people not to do `'<![CDATA[' . esc_xml(...) . ']]>'`

pbiron:
> > Also, `esc_xml()` should not do any modifications within CDATA Sections
> 
> How would you detect CDATA Sections? I'd say if someone uses them, they shouldn't be calling the function? We could suggest people not to do `'<![CDATA[' . esc_xml(...) . ']]>'`

that would be fine, if that's what they wanted.

The case I'm thinking of is plugin X does

```
update_post_meta(
   $post->ID,
   'my_meta_key',
   'This is a <![CDATA[test of the <emergency>]]> broadcast system'
);
```

and plugin Y (which adds custom properties to a sitemap, but doesn't have control over what plugin X wrote to post meta) does:
```
$url['some_custom_property'] = esc_xml( get_post_meta( $post->ID, 'my_meta_key', true ) );
```

swissspidy:
Ah yes, makes sense 👍 Hope it's not too tricky to implement this, otherwise we go down a rabbit hole trying not to cause security issues...

pbiron:
> Ah yes, makes sense 👍 Hope it's not too tricky to implement this, otherwise we go down a rabbit hole trying not to cause security issues...

Not hard, but it did take [ALOT](http://hyperboleandahalf.blogspot.com/2010/04/alot-is-better-than-you-at-everything.html) of experimentation to get the regex that separates out CDATA Sections both: 1) correct and 2) understandable by mere mortals :-)

inc/functions.php
```diff
@@ -117,3 +117,146 @@ function wp_sitemaps_get_max_urls( $object_type ) {
...
+if ( ! function_exists( 'esc_xml_non_cdata_section' ) ) :
+	/**
+	 * Escaping for non-CDATA Section XML blocks.
+	 *
+	 * @since 5.5.0
+	 *
+	 * @param string $text Text to escape.
+	 * @return string
+	 */
+	function esc_xml_non_cdata_section( $text ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
```
swissspidy: IMO this should be renamed to `_esc_xml_non_cdata_section` and marked with `@access private`,  as its intended for internal use only.

swissspidy (approved):
LGTM with one nit

inc/functions.php
pbiron: sounds good.

Will update shortly.

PR #193, opened 2020-05-26
Fix API oddity so that provider without subtypes can actually return empty array

### LINKED ISSUE #180: Rethink behavior in `Core_Sitemaps_Provider::get_object_subtypes()`
See discussion at https://github.com/GoogleChromeLabs/wp-sitemaps/pull/174/files#r423261977

felixarntz:
### Issue Number
Fixes #180 

### Description
Fixing this was very straightforward, we just needed to add a few lines of code to manually handle the empty case, which makes the code more understandable and allows a proper empty array return value for providers that don't support any object subtypes.

Tests still all pass, sitemaps and their names remain unchanged, so this should be good to go.

### Type of change

### Acceptance criteria


swissspidy (approved):
Neat!

inc/class-core-sitemaps-provider.php
```diff
@@ -110,9 +110,20 @@ public function get_sitemap_type_data() {
...
+			);
+			return $sitemap_data;
+		}
+
+		// Otherwise, include individual sitemaps for every object subtype.
 		foreach ( $object_subtypes as $object_subtype_name => $data ) {
 			$sitemap_data[] = array(
-				'name'   => $object_subtype_name,
+				'name'  => $object_subtype_name,
 				'pages' => $this->max_num_pages( $object_subtype_name ),
```
swissspidy: Looks like `$object_subtype_name` needs to be explicitly cast to a string?

```
Parameter #1 $object_subtype of method                              
         Core_Sitemaps_Provider::max_num_pages() expects string, int|string  
         given. 
```

felixarntz: Where would this be an integer? This should always be a string, if it's not, we may need to fix something elsewhere.

swissspidy: The static analysis (PHPStan) doesn't know that `$object_subtypes` is an associative array keyed with strings.

Not sure why it wasn't flagged before. But `$this->max_num_pages( (string) $object_subtype_name )` should do the trick.

felixarntz: Not familiar with that tool, seems a bit odd to me if you can't configure it like that - or it's configured overly strict here, cause this is a false positive. But anyway, let's do the string cast then.

PR #200, opened 2020-05-27
Refactor `assertXMLEqual()` to work around a bug in `DOMDocument::loadXml()`

pbiron:
### Description
The PHP Manual states that [DOMDocument::loadXml()](https://www.php.net/manual/en/domdocument.loadxml.php) "Returns TRUE on success or FALSE on failure.".  However, that's not 100% true.  If the XML is not-well-formed it _usually_ returns false.  But if it is not-well-formed because **a namespace prefix is used but not declared**, it incorrectly returns true.

Luckily, in all cases of non-well-formed XML (that I've checked), libxml_get_last_error() returns the error.  So, this refactors `Test_WP_Sitemaps_Renderer::loadXml()` (which is used by `Test_WP_Sitemaps_Renderer::assertXMLEqual()`) to check that instead of the return value from `DOMDocument::loadXml()`.

### Type of change

### Acceptance criteria

### DIFF
--- a/tests/phpunit/sitemaps-renderer.php
+++ b/tests/phpunit/sitemaps-renderer.php
@@ -234,10 +234,12 @@ class Test_WP_Sitemaps_Renderer extends WP_UnitTestCase {
 		libxml_clear_errors();
 
 		$xml_dom = new DOMDocument();
+		$xml_dom->loadXML( $xml, $options );
+		$libxml_last_error = libxml_get_last_error();
 
-		$this->assertTrue(
-			$xml_dom->loadXML( $xml, $options ),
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

swissspidy:
Works like a charm: https://travis-ci.com/github/GoogleChromeLabs/wp-sitemaps/jobs/340862305

pbiron:
> Works like a charm: https://travis-ci.com/github/GoogleChromeLabs/wp-sitemaps/jobs/340862305

Yup, that was the case that made me realize that `DOMDocument::loadXml()` was incorrectly returning `true`.

PR #201, opened 2020-05-31
Limit elements in sitemap entries to those defined in the Sitemaps spec.

pbiron:
### Description

#151 and #184 are attempts to provide support for sitemap extensions, but we have yet to come up with a "developer friendly" way of allowing **arbitrary** XML in extension namespaces (including XML attributes!); hence, this follow-up to https://github.com/GoogleChromeLabs/wp-sitemaps/pull/184#issuecomment-636055985.

This PR:

1. limits elements in sitemap entries to those defined in the Sitemaps spec: `<loc>`, `<lastmod>`, `<changefreq>`, `<priority>`
   * also fixes a bug (that I'd never noticed) to ensure that URLs are correctly escaped (previously, the value was escaped when the child name of `url` was `url`...it now correctly checks that the child name is `loc`)
2. updates the sitemap stylesheet to be able to display `<lastmod>`, `<changefreq>` and `<priority>` if they are present in the sitemap
   * a column for one of these elements does **not** appear in the rendered output if the element is not present in the sitemap
   * the columns are always in the order defined in the Sitemaps spec (no custom columns can be added, since no "custom" elements can be added to sitemaps)
   * also adds `html/@lang` (for i18n) and, `th/@class` and `td/@class` (to allow plugins to style the columns, via the exsting `get_stylesheet_css` filter)
3. cleans up the sitemap index stylesheet to be similar in XSLT "style" to the sitemap stylesheet
   * also adds `html/@lang` (for i18n) and, `th/@class` and `td/@class` (to allow plugins to style the columns, via the exsting `get_stylesheet_css` filter)
   * changes the string "This XML Sitemap contains %s URLs" to be "Number of URLs in this XML Sitemap: %s." (the same as for the provider sitemap stylesheet) to avoid i18n problems
4. updates the unit tests
   * `Test_WP_Sitemaps_Renderer::test_get_sitemap_xml_extra_attributes()` renamed to `Test_WP_Sitemaps_Renderer::test_get_sitemap_xml_extra_elements()`, to be more accurate about what it tests :-)
   * reverses the sense of the test.  previously it tested that extra elements were added; now it tests that only elements defined in the Sitemaps spec are added

This PR does **NOT** do any other validation on the generate sitemaps (e.g., all entries contain a `<loc>`,  that the children of `<url>` are in the order defined in the Sitemaps spec, that the value of `<changefreq>` is one of the values enumerated in the Sitemaps spec, etc).  I'll open a separate issue about that.

### Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Steps to test

To test that elements other than `<loc>`, `<lastmod>`, `<changefreq>`, `<priority>` are not added to the sitemap do something like:

```
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry ) {
	$sitemap_entry['custom'] = 'custon';

	return $sitemap_entry;
} );
```
and check the resulting XML to verify that no `<custom>` children exist.

To test that additional columns are added to the output rendered by the stylesheet do something like:

```
// see that 'Last Modified' column is rendered when 'lastmod' children appear in the sitemap.
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry ) {
	$sitemap_entry['lastmod'] = 'lastmod';

	return $sitemap_entry;
} );

// see that 'Last Modified' and 'Priority' columns are rendered when 'lastmod' and 'priority'
// children appear in the sitemap.
// also see that columns are in the order defined in the Sitemaps spec even when they are
// added to the entry in a different order.
add_filter( 'wp_sitemaps_taxonomies_entry', function( $sitemap_entry ) {
	$sitemap_entry['priority'] = 'priority';
	$sitemap_entry['lastmod']  = 'lastmod';

	return $sitemap_entry;
} );

// see that values in the table body rows are in the correct columns when an element is added
// to only some entries.
add_filter( 'wp_sitemaps_users_entry', function( $sitemap_entry ) {
	// only add 'lastmod' to some entries.
	if ( random_int( 0, 1 ) ) {
		$sitemap_entry['lastmod'] = 'lastmod';
	}
	$sitemap_entry['priority'] = 'priority';

	return $sitemap_entry;
} );
```

### Acceptance criteria
- [X] If the changes are visual, I have cross browser / device tested.

### DIFF
--- a/inc/class-wp-sitemaps-renderer.php
+++ b/inc/class-wp-sitemaps-renderer.php
@@ -209,7 +209,7 @@ class WP_Sitemaps_Renderer {
 						__METHOD__,
 						/* translators: %s: list of element names */
 						sprintf(
-							__( 'Fields other than %s are not currently supported for sitemaps', 'core-sitemaps' ),
+							__( 'Fields other than %s are not currently supported for sitemaps.', 'core-sitemaps' ),
 							implode( ',', array( 'loc', 'lastmod', 'changefreq', 'priority' ) )
 						),
 						'5.5.0'
--- a/inc/class-wp-sitemaps-renderer.php
+++ b/inc/class-wp-sitemaps-renderer.php
@@ -198,12 +198,12 @@ class WP_Sitemaps_Renderer {
 		foreach ( $url_list as $url_item ) {
 			$url = $urlset->addChild( 'url' );
 
-			// Add each attribute as a child node to the URL entry.
-			foreach ( $url_item as $attr => $value ) {
-				if ( 'url' === $attr ) {
-					$url->addChild( $attr, esc_url( $value ) );
-				} else {
-					$url->addChild( $attr, esc_attr( $value ) );
+			// Add each element as a child node to the URL entry.
+			foreach ( $url_item as $name => $value ) {
+				if ( 'loc' === $name ) {
+					$url->addChild( $name, esc_url( $value ) );
+				} elseif ( in_array( $name, array( 'lastmod', 'changefreq', 'priority' ), true ) ) {
+					$url->addChild( $name, esc_attr( $value ) );
 				}
 			}
 		}
--- a/inc/class-wp-sitemaps-renderer.php
+++ b/inc/class-wp-sitemaps-renderer.php
@@ -204,6 +204,16 @@ class WP_Sitemaps_Renderer {
 					$url->addChild( $name, esc_url( $value ) );
 				} elseif ( in_array( $name, array( 'lastmod', 'changefreq', 'priority' ), true ) ) {
 					$url->addChild( $name, esc_attr( $value ) );
+				} else {
+					_doing_it_wrong(
+						__METHOD__,
+						/* translators: %s: list of element names */
+						sprintf(
+							__( 'Fields other than %s are not currently supported for sitemaps', 'core-sitemaps' ),
+							implode( ',', array( 'loc', 'lastmod', 'changefreq', 'priority' ) )
+						),
+						'5.5.0'
+					);
 				}
 			}
 		}
--- a/inc/class-wp-sitemaps-stylesheet.php
+++ b/inc/class-wp-sitemaps-stylesheet.php
@@ -52,62 +52,85 @@ class WP_Sitemaps_Stylesheet {
 		$text        = sprintf(
 			/* translators: %s: number of URLs. */
 			__( 'Number of URLs in this XML Sitemap: %s.', 'core-sitemaps' ),
-			'<xsl:value-of select="count(sitemap:urlset/sitemap:url)"/>'
+			'<xsl:value-of select="count( sitemap:urlset/sitemap:url )" />'
 		);
 
-		$url = esc_html__( 'URL', 'core-sitemaps' );
+		$lang       = get_language_attributes( 'html' );
+		$url        = esc_html__( 'URL', 'core-sitemaps' );
+		$lastmod    = esc_html__( 'Last Modified', 'core-sitemaps' );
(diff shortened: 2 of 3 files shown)


pbiron:
Yesterday I was playing with some alternate means of allowing extension namespace elements (similar to https://github.com/GoogleChromeLabs/wp-sitemaps/pull/184#issuecomment-636033248).

Since each of the children of `<url>` defined in the Sitemaps spec can only occur once, using their element names as keys in the `$sitemap_entry` array makes sense, e.g.,

```
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry, $post, $post_type ) {
	$sitemap_entry['lastmod'] = mysql2date( DATE_W3C, $post->post_modified_gmt, false );

	return $sitemap_entry;
}
```

But in order for plugins to (eventually) support Image, Video, and [Localized](https://support.google.com/webmasters/answer/189077?hl=en) sitemaps they will need to be able to add the same child element multiple times.  Thus, regardless of whatever representation we eventually come up with  for extension elements, it is likely to be something like:

```
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry, $post, $post_type ) {
	$sitemap_entry[] = array(
		array(
			// add 1st image:image element
		),
		array(
			// add another image:image element
		),
	);

	return $sitemap_entry;
}
```

It will be no problem to eventually modify `WP_Sitemaps_Renderer::get_sitemap_xml()` to handle both alpha and numeric keys in a `$sitemap_entry`, but I just wanted to point out the possible disconnect for developers if they want to add both `<lastmod>` and `<image:image>`, as in:

```
add_filter( 'wp_sitemaps_posts_entry', function( $sitemap_entry, $post, $post_type ) {
	$sitemap_entry['lastmod'] = mysql2date( DATE_W3C, $post->post_modified_gmt, false );
	$sitemap_entry[]          = array(
		array(
			// add 1st image:image element
		),
		array(
			// add another image:image element
		),
	);

	return $sitemap_entry;
}
```

Again, don't think that should be considered a blocker, just wanted to point out the forward compatibility issue.

swissspidy (approved):
I like this!

swissspidy (approved):

inc/class-wp-sitemaps-renderer.php
```diff
@@ -198,12 +198,12 @@ public function get_sitemap_xml( $url_list ) {
...
-					$url->addChild( $attr, esc_url( $value ) );
-				} else {
-					$url->addChild( $attr, esc_attr( $value ) );
+			// Add each element as a child node to the URL entry.
+			foreach ( $url_item as $name => $value ) {
+				if ( 'loc' === $name ) {
+					$url->addChild( $name, esc_url( $value ) );
+				} elseif ( in_array( $name, array( 'lastmod', 'changefreq', 'priority' ), true ) ) {
+					$url->addChild( $name, esc_attr( $value ) );
 				}
```
swissspidy, line 217: @pbiron to answer your question about `_doing_it_wrong`, I think if the wording is right we can add it.

e.g.

```php

// ...
} else {
	_doing_it_wrong(
		__FUNCTION__,
		/* translators: %s: list of element names */
		sprintf(
			__( 'Fields other than %s are not currently supported for sitemaps' ),
			implode( ',', array( 'loc', 'lastmod', 'changefreq', 'priority' ) )
		)
	)
}
```

pbiron: that's just what I was thinking.

Will there be any problem with that `_doing_it_wrong()` being removed in a future version when we do support extension elements?

swissspidy: Not really

pbiron: will update the PR after the meeting

PR #203, opened 2020-06-03
XML escape strings

pbiron:
### Description

Now that #192 has been merged, use `esc_xml()` to escape XML content.

### Type of change

### Acceptance criteria


inc/class-wp-sitemaps-stylesheet.php
```diff
@@ -180,7 +191,7 @@ public function get_sitemap_index_stylesheet() {
 	  Set variables for whether lastmod occurs for any sitemap in the index.
 	  We do this up front because it can be expensive in a large sitemap.
 	  -->
-	<xsl:variable name="has-lastmod"    select="count( /sitemap:sitemapindex/sitemap:sitemap/sitemap:lastmod )"    />
+	<xsl:variable name="has-lastmod" select="count( /sitemap:sitemapindex/sitemap:sitemap/sitemap:lastmod )"    />
```
swissspidy:
```suggestion
	<xsl:variable name="has-lastmod" select="count( /sitemap:sitemapindex/sitemap:sitemap/sitemap:lastmod )" />
```

inc/class-wp-sitemaps-stylesheet.php
```diff
@@ -42,24 +42,29 @@ public function render_stylesheet( $type ) {
...
 			/* translators: %s: URL to sitemaps documentation. */
-			__( 'This XML Sitemap is generated by WordPress to make your content more visible for search engines. Learn more about XML sitemaps on <a href="%s">sitemaps.org</a>.', 'core-sitemaps' ),
-			__( 'https://www.sitemaps.org/', 'core-sitemaps' )
+			'<a href="%s">sitemaps.org</a>',
+			esc_url( __( 'https://www.sitemaps.org/', 'core-sitemaps' ) )
 		);
-		$text        = sprintf(
+		$description   = sprintf(
+			/* translators: %s: link to sitemaps documentation. */
+			esc_xml( __( 'This XML Sitemap is generated by WordPress to make your content more visible for search engines. Learn more about XML sitemaps on %s.', 'core-sitemaps' ) ),
```
swissspidy, line 54: Hmm, having `<a href="%s">some text</a>` in translatable strings is a standard WordPress core practice. Why the change here?

inc/class-wp-sitemaps-stylesheet.php
```diff
@@ -42,24 +42,29 @@ public function render_stylesheet( $type ) {
 	 * @since 5.5.0
 	 */
 	public function get_sitemap_stylesheet() {
-		$css         = $this->get_stylesheet_css();
-		$title       = esc_html__( 'XML Sitemap', 'core-sitemaps' );
-		$description = sprintf(
+		$css           = $this->get_stylesheet_css();
+		$title         = esc_xml( __( 'XML Sitemap', 'core-sitemaps' ) );
+		$sitemaps_link = sprintf(
```
pbiron, line 47: @swissspidy I had to pull the `<a>` tag out of the translatable string.  Can you please verify I did that correctly?  Similarly, on line159.

inc/class-wp-sitemaps-stylesheet.php
pbiron: Because `esc_xml()` would escape the `<` and `>` chars to `&lt;` and `&gt;`.  The same thing would happen with:

```
$foo = sprintf(
	esc_html__( 'This is a <a href="%s">test</a>', 'domain' ),
	esc_url( __( 'https://example.com', 'domain' ) )
);
```

swissspidy: Perhaps we shouldn't be using `esc_*` then, just like we aren't right now (i.e., revert this change)? Translations come from a trusted source anyway.

pbiron: I don't have a lot of experience with setting things up for translation (as you saw from my comments on the plural issue :-), so I'll let you make the call here.

But, if we include markup in the translatable string, then I think we should use `_x()` instead of `__()`, where the `$context` param tells translators to make sure that the translations are properly XML escaped (e.g., no HTML named character references)...in the **very** unlikely event that string is used in an HTML context.

The whole reason for `esc_xml()` is because the rules of XML are so much stricter than those for HTML, and the vast majority of people aren't familiar with those strict rules, and think that if they can do something in HTML they can do it in XML because "they are both pointy-bracket syntaxes".

And even when the HTML spec says "Thou shalt not do _this_" e.g., (see, https://www.w3.org/TR/html52/syntax.html#elements-attributes and https://www.w3.org/TR/html52/syntax.html#ambiguous-ampersand)

```
<a href='https://example.com'
	title='this is technically illegal because &unknwon; is what the HTML5 spec refers to as an &quot;ambiguous ampersand&quot;'>test</a>
```

browser vendors know that most HTML authors often get things wrong and so browsers accept all kinds of things that are technically illegal.

The same is **not** true of XML processors (including the XSLT processors built into all modern browsers).  So, if the translation of that string were to contain `&rsaquo;` or `&unknown;` the rendered version of the sitemap would be a white screen of death...and if you loaded the stylesheet containing those character references in a browser you get a screen such as
stylesheet-error

So, again, I'll let you make the call on this one.

The last thing I'll say is to point out that there are hundreds of uses of `esc_html__()` (and kin)  in core, and couldn't they all be replaced by `__()` since "Translations come from a trusted source anyway"?

swissspidy: Thanks for this very thorough reply! Really appreciate it.

Let's stick with this then :-)

swissspidy (approved):

Issue #164, opened 2020-04-28
Road towards WordPress Core


swissspidy:
This thread aims to cover things needed to pave the way for an eventual merge proposal into WordPress core.

Sub-issues will be created as-needed.

The tasks are more or less in desired chronological order:

* [x] Release version 0.3.0. Including these PRs:
  * [x] #138 - Private sites
  * [x] #150 - Improving rewrite rules
* [x] Land other outstanding PRS, perhaps for a 0.4.0 release. At the moment there are:
  * [x] #163 - Smarter stylesheet
  * [x] #124 - Test coverage for filters
* [x] Get issue count to zero.
* [x] Update all inline documentation to adhere to WordPress coding standards: #173
* [x] Refactor code base
  * [x] Use core conventions around object type / object subtype names and docs: #174 
  * [x] Drop `core_sitemaps` prefix in functions, hooks, and class file names (In progress).
  * [x] Bail early in `core-sitemaps.php` if functionality is already built into core (e.g. if functions already exist). (in progress)
* [ ] Get audit from WordPress security team
* [x] Cover everything under https://make.wordpress.org/core/handbook/about/release-cycle/features-as-plugins/
* [x] Create ticket on WordPress Trac - Complete: https://core.trac.wordpress.org/ticket/50117
* [x] Create pull request towards https://github.com/WordPress/wordpress-develop/ with the sitemaps code & tests
* [x] Write merge proposal post

adamsilverstein:
@swissspidy I will take a pass at "Update all inline documentation to adhere to WordPress coding standards" - is there anything in particular you noticed that needs adjusting? I see that `phpcs` is in place and enforcing having inline docs :). 

I do see that the language for descriptions is a bit off - the documentation standards call out using [third-person singular verbs](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#language) which we do not - I will start with that.

felixarntz:
@adamsilverstein The 3P singular verbs was I was thinking about too here. We should also consider updating the `since` annotations - many are missing, and the ones are there refer to plugin versions. Maybe we should already set these to 5.5 instead, or something that's easy to replace with the exact WP version later?

adamsilverstein:
@felixarntz Good idea, I'll set them all to `@since 5.5.0` since that is our target for merge. I spotted a few other irregularities along the way, mostly though, this code is very well documented!

felixarntz:
[...]

PR #204, opened 2020-06-04
Prepare for 0.4.0 release

swissspidy:
### Issue Number

See #164

### Description

Numerous changes in this PR:

* Bumps
* Updates changelog
* Updates documentation
* Renames 1 filter for consistency
* Updates usage of `esc_xml` (introduced in #203)
* Adds type hints (as suggested upstream in https://github.com/WordPress/wordpress-develop/pull/310)
* Removes `esc_xml` l10n helpers (as suggested in https://github.com/WordPress/wordpress-develop/pull/310)

### Type of change

### Acceptance criteria


inc/class-wp-sitemaps-renderer.php
```diff
@@ -255,11 +255,11 @@ static function () {
 
 			wp_die(
 				sprintf(
-				/* translators: %s: SimpleXML */
-					esc_xml( __( 'Could not generate XML sitemap due to missing %s extension', 'core-sitemaps' ) ),
+					/* translators: %s: SimpleXML */
+					__( 'Could not generate XML sitemap due to missing %s extension', 'core-sitemaps' ),
 					'SimpleXML'
 				),
-				esc_xml( __( 'WordPress &rsaquo; Error', 'core-sitemaps' ) ),
```
pbiron: This one does need to call `esc_xml()`, so that `&rsaquo;` gets escaped to `›`.  The other one _should_ be XML escaped, in case there are HTML entities in a translation.

FYI: when the core merge happens, `_xml_wp_die_handler()` should be rewritten to call `esc_xml()`, instead of `htmlspecialchars()` on both the `$message` and `$title` params...and to **not** wrap those in CDATA sections in the output; and when happens these can both be changed back to `__()`.

swissspidy: Just so I get this right, the `esc_xml` is still needed, despite the CDATA sections in `_xml_wp_die_handler`?

pbiron (changes requested):
Added a comment about use of `esc_xml()` when calling `wp_die()`.

Other than that, LGTM.

inc/class-wp-sitemaps-renderer.php
pbiron: yes...so that HTML entities get converted to their code points.

swissspidy: Got it, thanks! Guess I was a bit over-excited about removing potentially unneeded code. But it all makes sense now :-)

pbiron: BTW, the `$learn_more` link is a much cleaner way to handle that in the stylesheets!

You educate me about l10n and I educate you about XML :-)

swissspidy:
@pbiron Reverted the change as discussed. Mind reviewing again? :-)

pbiron (approved):

